PR #76, opened 2026-01-03
feat(vector): Vectorize separation() to support array inputs

### LINKED ISSUE #75: Vectorize separation() to support array inputs
## Summary

The `separation()` function in `conops/common/vector.py` currently only supports scalar RA/Dec inputs. This causes `SolarPanelSet.illumination_and_power()` to fail when called with an array of times.

## Problem

When calling `illumination_and_power()` with multiple times, the underlying `separation()` call receives array RA/Dec values but can only handle scalars. This results in incorrect calculations or errors.

There is currently a skipped test (`test_illumination_and_power_with_array_times`) that documents this limitation.

## Proposed Solution

1. Update `radec2vec()` to handle both scalar and array inputs
2. Update `separation()` to use vectorized operations for arrays while preserving scalar behavior
3. Add `float()` casts in callers that expect scalar returns (for mypy compatibility)
4. Enable the skipped test

## Acceptance Criteria

- [ ] `separation()` returns scalar for scalar inputs, array for array inputs
- [ ] `SolarPanelSet.illumination_and_power()` works with array time inputs
- [ ] `test_illumination_and_power_with_array_times` passes (no longer skipped)
- [ ] All existing tests continue to pass

Tohuvavohu:
## Summary

- Vectorize `radec2vec()` and `separation()` to handle both scalar and array RA/Dec inputs
- Enable batch angular distance calculations for `SolarPanelSet.illumination_and_power()`
- Enable previously skipped test

Fixes #75

## Changes

| File | Change |
|------|--------|
| `conops/common/vector.py` | `radec2vec()` returns (n, 3) for arrays, (3,) for scalars |
| `conops/common/vector.py` | `separation()` uses vectorized path for arrays |
| `conops/config/acs.py` | Add `float()` cast for mypy compatibility |
| `conops/simulation/passes.py` | Add `float()` cast for mypy compatibility |
| `tests/solar_panel/test_solar_panel_comprehensive.py` | Remove `@pytest.mark.skip` decorator |

## Test plan

- [x] All 1480 tests pass
- [x] `test_illumination_and_power_with_array_times` now runs and passes
- [x] mypy strict mode passes

### DIFF
--- a/conops/common/vector.py
+++ b/conops/common/vector.py
@@ -3,14 +3,30 @@ import numpy.typing as npt
 from pyproj import Geod
 
 
-def radec2vec(ra: float, dec: float) -> npt.NDArray[np.float64]:
-    """Convert RA/Dec angle (in radians) to a vector"""
+def radec2vec(
+    ra: float | npt.NDArray[np.float64], dec: float | npt.NDArray[np.float64]
+) -> npt.NDArray[np.float64]:
+    """Convert RA/Dec angle (in radians) to a unit vector.
 
-    v1 = np.cos(dec) * np.cos(ra)
-    v2 = np.cos(dec) * np.sin(ra)
-    v3 = np.sin(dec)
+    Args:
+        ra: Right ascension in radians (scalar or array)
+        dec: Declination in radians (scalar or array)
 
-    return np.array([v1, v2, v3], dtype=np.float64)
+    Returns:
+        Unit vector(s). Shape (3,) for scalar input, (n, 3) for array input.
+    """
+    ra_arr = np.atleast_1d(ra)
+    dec_arr = np.atleast_1d(dec)
+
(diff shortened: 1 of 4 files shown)


jak574 (changes requested):
Did you forget to check in a file? I don't see any code that uses the vectorized separation?

Tohuvavohu:
Closing this PR after analysis shows the vectorized `separation()` isn't worth the added complexity:

- The main use case (emergency charging search) only calls `angular_separation` when `max_slew_deg` is configured, which is `None` by default
- Even with slew limits enabled, max savings would be ~0.45s per 12h DITL (~15%), but early-exit behavior reduces actual benefit
- The 123x micro-benchmark speedup doesn't translate to meaningful real-world improvement

Per discussion with @jak574: removing the unused capability rather than adding complexity for a speculative use case. Will delete the skipped test separately.

PR #78, opened 2026-01-03
fix(vector): Use 180° threshold in roll_over_angle for shortest path

### LINKED ISSUE #77: roll_over_angle() takes long path for RA jumps > 180°
## Summary

The `roll_over_angle()` function uses a 300° threshold to detect angle wrap-arounds. This causes incorrect interpolation when RA changes by amounts between 180° and 300°.

## Problem

Near celestial poles, RA can legitimately change by large amounts (e.g., 285°) in a short time during slews. The current 300° threshold fails to detect this as a wrap-around, causing:

1. Linear interpolation takes the "long way" around (285° path instead of 75° short path)
2. This results in pointing rate violations during slews near the poles
3. Slew paths are incorrect, potentially missing targets

## Example

```python
# RA path near pole: 350° → 75° (285° apparent change)
angles = [350.0, 75.0]
result = roll_over_angle(angles)

# Current behavior: result = [350, 75], diff = -275° (WRONG - long path)
# Expected behavior: result = [350, 435], diff = +85° (CORRECT - short path)
```

## Root Cause

The threshold of 300° is too high. Any angle change should take the shortest path, which means the threshold should be 180°.

## Proposed Fix

Change threshold from 300° to 180° in `roll_over_angle()`.

Tohuvavohu:
## Summary

Fix `roll_over_angle()` to always take the shortest path around the circle by using a 180° threshold instead of 300°.

Fixes #77

## Problem

Near celestial poles, RA can change by ~285° during slews. The old 300° threshold failed to detect this as a wrap-around, causing interpolation to take the "long way" (285° instead of 75°), resulting in pointing rate violations.

## Changes

| Change | Before | After |
|--------|--------|-------|
| Threshold | 300° | 180° |
| Flip accumulation | `flip = ±360` | `flip ±= 360` |

## Test plan

- [x] New test `test_roll_over_angle_large_jump_near_poles` verifies fix
- [x] All 1480 existing tests pass
- [x] mypy strict mode passes

### DIFF
--- a/conops/common/vector.py
+++ b/conops/common/vector.py
@@ -124,7 +124,12 @@ def great_circle(
 def roll_over_angle(
     angles: npt.NDArray[np.float64] | list[float],
 ) -> npt.NDArray[np.float64]:
-    """Make a list of angles that include a roll over (e.g. 359.9 - 0.1) into a smooth distribution"""
+    """Make a list of angles that include a roll over (e.g. 359.9 - 0.1) into a smooth distribution.
+
+    Uses 180° threshold to ensure interpolation always takes the shortest path
+    around the circle. This is critical for slew paths that cross near the poles
+    where RA can change rapidly.
+    """
     outangles = list()
     last = -1.0
     flip = 0.0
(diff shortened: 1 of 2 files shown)


jak574 (approved):
lgtm, good catch.

PR #85, opened 2026-01-03
perf: Precompute eclipse status at simulation start

Tohuvavohu:
## Summary

- Batch-compute eclipse status for all simulation timesteps at initialization using `in_constraint_batch()` API
- Eclipse depends only on orbital position (not pointing direction), so can be precomputed once
- Adds `first_constraint_violation()` method for batch constraint checking (enables future optimizations)

**Performance impact:** ~0.18s savings (5.4%) on 12-hour DITL simulation

## Changes

- `conops/config/constraint.py`: Add `precompute_eclipse()` and `first_constraint_violation()` methods; modify `in_eclipse()` to check precomputed cache first
- `conops/ditl/queue_ditl.py`: Call `precompute_eclipse()` during simulation setup

## Test plan

- [x] All existing DITL tests pass
- [x] mypy strict type checking passes
- [x] Verified 5.4% performance improvement via profiling

### DIFF
--- a/conops/config/constraint.py
+++ b/conops/config/constraint.py
@@ -183,10 +183,51 @@ class Constraint(BaseModel):
 
     def in_eclipse(self, ra: float, dec: float, time: float) -> bool:
         assert self.ephem is not None, "Ephemeris must be set to use in_eclipse method"
+        # Check precomputed cache first (keyed by rounded time)
+        if hasattr(self, "_eclipse_precomputed") and self._eclipse_precomputed:
+            time_key = int(time)  # Round to nearest second
+            if time_key in self._eclipse_precomputed:
+                return self._eclipse_precomputed[time_key]
         # Eclipse constraint is special - create once and cache
         if not hasattr(self, "_eclipse_constraint"):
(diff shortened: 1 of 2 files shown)


conops/config/constraint.py
```diff
@@ -183,11 +183,52 @@ def in_earth(self, ra: float, dec: float, time: float) -> bool:
...
+
+        # Ensure eclipse constraint exists
+        if not hasattr(self, "_eclipse_constraint"):
+            self._eclipse_constraint = rust_ephem.EclipseConstraint()
+
+        # Convert to datetime objects for batch API
+        times = [dtutcfromtimestamp(t) for t in utimes]
+
+        # Batch compute - use dummy RA/Dec since eclipse is position-only
+        result = self._eclipse_constraint.in_constraint_batch(
```
jak574: No need to use `in_constraint_batch` for eclipse, that is meant to batch calculate constraints for large arrays or RA/Dec, for EclipseConstraint RA/Dec is basically ignored, and just there so that the arguments for that constraint match others. 

TL;DR: Just use `in_constraint`.

PR #87, opened 2026-01-03
Increase slew path resolution from 20 to 100 steps

Tohuvavohu:
## Summary

- Increases slew path resolution from 20 to 100 steps in `predict_slew()`
- Adds tests quantifying interpolation error for different step counts

## Problem

Near celestial poles, linear RA/Dec interpolation between coarse path points deviates from the true great-circle path. With only 20 steps, this deviation can cause apparent slew rates to exceed the configured maximum, triggering false rate validation failures.

## Measurements

| Steps | Max Deviation | Runtime (per path) |
|-------|---------------|-------------------|
| 20    | 0.27°         | ~12µs            |
| 100   | 0.055°        | ~30µs            |

- Improvement: 4.8x reduction in interpolation error
- Runtime cost: 2.5x slower, but still negligible (~30ms total for 1000 slews)

## Test Plan

- [x] New tests in `TestSlewPathResolution` verify deviation thresholds
- [x] All existing slew tests pass
- [x] Runtime cost test confirms acceptable performance

### DIFF
--- a/tests/slew/test_slew.py
+++ b/tests/slew/test_slew.py
@@ -251,3 +251,114 @@ class TestPredictSlew:
         slew, ra_path, dec_path = slew_predict_setup
         slew.predict_slew()
         assert np.allclose(slew.slewpath[1], dec_path)
+
+
+class TestSlewPathResolution:
+    """Tests demonstrating why 100 steps is better than 20 for slew paths.
+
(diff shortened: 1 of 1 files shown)


jak574 (approved):
Lgtm

PR #136, opened 2026-05-13
fix: under-collected science plan entries

Tohuvavohu:
## Summary

Closes #135.

This fixes the scheduler invariant that science `AT` entries should only be emitted when they collect at least `ss_min` after slew/INSAA.

Changes:
- reject science targets that cannot collect `ss_min` before the next science deadline, then retry the next queue candidate
- drop science plan fragments that get closed before collecting `ss_min`
- avoid normal science termination while ACS is still slewing
- keep interrupted under-collected science targets retryable after emergency charging
- clamp `PlanEntry.exposure` at zero so serialized plans do not report negative exposure

## Validation

- `ruff check conops/ditl/queue_ditl.py conops/targets/plan_entry.py tests/scheduler_queue/test_queue_ditl.py tests/plan_entry/test_plan_entry.py`
- `mypy --strict conops/ditl/queue_ditl.py conops/targets/plan_entry.py`
- `pytest`

Result: `1920 passed, 1 skipped`.

Closed-loop regenerated comparison against the reconstructed problematic Tamalpais case:
- invalid v0: 36 short science entries, 16 negative exposures
- fixed run: 0 short science entries, 0 negative exposures
- observing time increased from `18:30:42` to `19:33:02`
- slewing decreased from `04:35:18` to `03:30:43`

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -19,6 +19,7 @@ from ..common.vector import attitude_to_quat
 from ..config import MissionConfig
 from ..simulation.acs_command import ACSCommand
 from ..simulation.emergency_charging import EmergencyCharging
+from ..simulation.passes import pass_slew_trigger_buffer
 from ..simulation.roll import optimum_roll
 from ..simulation.slew import Slew
 from ..targets import Plan, Pointing, Queue
@@ -447,7 +448,7 @@ class QueueDITL(DITLMixin, DITLStats):
 
         # Make sure the last PPT of the day ends (if any)
         if self.plan:
-            self.plan[-1].end = utime
+            self._close_last_plan_entry(utime)
 
         return True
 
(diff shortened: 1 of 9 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -999,6 +1076,156 @@ def _get_constraint_name(
...
+        pass_deadline = self._next_pass_science_deadline(slew_end)
+        if pass_deadline is not None:
+            deadlines.append((pass_deadline, "pass"))
+
+        if not deadlines:
+            return None, None
+        return min(deadlines, key=lambda item: item[0])
+
+    def _can_retry_without_current_ppt(self) -> bool:
+        targets = getattr(self.queue, "targets", None)
```
jak574: Generally not a fan of using getattr, if the intent here is to return a `None`
if self.queue.targets is `[]` then do:
```suggestion
        targets = self.queue.targets or None
```

conops/ditl/queue_ditl.py
```diff
@@ -999,6 +1076,156 @@ def _get_constraint_name(
...
+            self._fetch_new_ppt(utime, ra, dec)
+        finally:
+            rejected_ppt.done = was_done
+
+    def _reject_current_ppt_if_insufficient_collect_time(
+        self, slew_end: float, utime: float, ra: float, dec: float
+    ) -> bool:
+        assert self.ppt is not None
+        ss_min = self._float_or_none(getattr(self.ppt, "ss_min", None))
+        if ss_min is None or ss_min <= 0:
```
jak574: What's the deal with this, it seems super defensive programming?

conops/ditl/queue_ditl.py
```diff
@@ -518,6 +519,73 @@ def _handle_fault_management(
...
+            return collection_time <= 0.0
+        return collection_time < max(0.0, ss_min)
+
+    def _pop_last_plan_entry(self) -> None:
+        entries = getattr(self.plan, "entries", None)
+        if entries is not None:
+            entries.pop()
+            return
+        pop = getattr(self.plan, "pop")
+        pop()
```
jak574: What? This looks like some weird vibe coding BS. 

```
self.plan.pop()
```

conops/ditl/queue_ditl.py
```diff
@@ -518,6 +519,73 @@ def _handle_fault_management(
...
+            return False
+
+        ss_min = cls._float_or_none(getattr(entry, "ss_min", None))
+        if ss_min is None:
+            return collection_time <= 0.0
+        return collection_time < max(0.0, ss_min)
+
+    def _pop_last_plan_entry(self) -> None:
+        entries = getattr(self.plan, "entries", None)
+        if entries is not None:
```
jak574: Again with the getattr

conops/ditl/queue_ditl.py
```diff
@@ -999,6 +1076,156 @@ def _get_constraint_name(
...
+        except (TypeError, ValueError):
+            return None
+
+    def _simulation_end_deadline(self) -> float | None:
+        if getattr(self, "uend", 0.0):
+            return float(self.uend)
+        try:
+            return float(self.end.timestamp())
+        except Exception:
+            return None
```
jak574: What is this doing? Why the getattr here? Is there ever a situation where
self.uend is None? It's set at the start of the simulation.

conops/ditl/queue_ditl.py
```diff
@@ -518,6 +519,73 @@ def _handle_fault_management(
...
+        begin = cls._float_or_none(getattr(entry, "begin", None))
+        raw_end = getattr(entry, "end", None) if end_time is None else end_time
+        end = cls._float_or_none(raw_end)
+        slewtime = cls._float_or_none(getattr(entry, "slewtime", None))
+        insaa = cls._float_or_none(getattr(entry, "insaa", 0.0))
+        if begin is None or end is None or slewtime is None:
+            return None
+        if insaa is None:
+            insaa = 0.0
+        return end - begin - max(0.0, slewtime) - max(0.0, insaa)
```
jak574: Generally not a fan of the super defensive programming here. _float_or_none and
getattr use specifically.

conops/ditl/queue_ditl.py
```diff
@@ -999,6 +1076,156 @@ def _get_constraint_name(
             return "ST Soft"
         return "Unknown"
 
+    @staticmethod
+    def _float_or_none(value: Any) -> float | None:
+        try:
+            return float(value)
+        except (TypeError, ValueError):
+            return None
+
```
jak574: Not a fan of this. Why is it needed?

jak574 (changes requested):
So, I have a lot of issues with this code, mostly because of a bunch of patterns I'm not a fan of and are overly defensive. However, I'm just going to go through them and rewrite myself to be cleaner.

jak574 (approved):
OK. So almost all my changes were due to the overly defensive coding practices. Maybe the root cause of this were getting unit tests to pass? I dunno, but it was very hard to read and made the code complex. Anyway, I went through and cleaned it all up, then fixed the unit tests that were relying on the defensive code. As far as I can tell, I did not change the intent of the code, which seems to work as expected in my tests. I approve, but you should check I didn't break anything.

PR #137, opened 2026-05-16
fix: plotly timeline hover start times

Tohuvavohu:
## Summary

- render Plotly DITL timeline hover start/duration values as concrete strings
- stop using `%{base}` / `%{x}` for timeline hover text, which can render as `-h`
- add regression coverage for resolved hover times

## Tests

- `PATH=".venv/bin:$PATH" pre-commit run --all-files`
- `.venv/bin/python -m pytest tests/visualization/test_ditl_timeline.py`


jak574 (approved):
lgtm

PR #139, opened 2026-05-16
feat: represent commanded ground-station passes as plan activities

Tohuvavohu:
## Summary
- Add `GSP` plan entries for ground-station pass activities that COAST actually commands
- Include pass metadata in plan JSON: `station`, `contact_begin`, `contact_end`
- Reserve the GSP window in the queue plan so science/charging entries are closed before the pass
- Avoid exporting overlapping unselected pass opportunities
- End the commanded pass based on `acs.current_pass`, so overlapping pass opportunities do not create accidental tail contacts
- Skip pass handling/export while ACS is in SAFE
- Exclude `GSP` entries from science-observation timeline extraction

Closes #138

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -19,7 +19,7 @@ from ..common.vector import attitude_to_quat
 from ..config import MissionConfig
 from ..simulation.acs_command import ACSCommand
 from ..simulation.emergency_charging import EmergencyCharging
-from ..simulation.passes import pass_slew_trigger_buffer
+from ..simulation.passes import Pass, pass_slew_trigger_buffer
 from ..simulation.roll import optimum_roll
 from ..simulation.slew import Slew
 from ..targets import Plan, PlanEntry, Pointing, Queue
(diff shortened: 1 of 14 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -797,12 +801,133 @@ def _schedule_groundstation_passes(self) -> None:
...
+        return True
+
+    def _record_gsp_plan_entry(self, gspass: Pass, reserved_begin: float) -> bool:
+        """Record an accepted ground-station pass command in the exported plan."""
+        key = self._ground_pass_key(gspass)
+        if key in self._planned_gsp_keys:
+            return True
+
+        station, pass_begin, pass_end = key
+        entry_begin = float(reserved_begin)
```
jak574: This is always float according to the type hinting, so why do we need a conversion here? Can `entry_begin` just be replaced with `reserved_begin ` throughout this method?

conops/ditl/queue_ditl.py
```diff
@@ -797,12 +801,133 @@ def _schedule_groundstation_passes(self) -> None:
...
+                self._close_last_plan_entry(entry_begin)
+
+        entry = PlanEntry(config=self.config)
+        entry.name = f"{station}_PASS"
+        entry.ra = float(gspass.gsstartra)
+        entry.dec = float(gspass.gsstartdec)
+        entry.begin = entry_begin
+        entry.end = pass_end
+        entry.slewtime = max(0, int(round(pass_begin - entry_begin)))
+        entry.obsid = int(gspass.obsid)
```
jak574: `obsid` here is defined as an `int`. Why the conversion?

jak574 (changes requested):
Couple of minor fixes, otherwise this looks good to me.

jak574 (approved):
LGTM, but please update documentation with API changes. 

I pushed a commit that does this, so good for merge.

PR #144, opened 2026-05-18
fix: cancel superseded ACS slew commands

Tohuvavohu:
Closes #141.

Summary:
- Treat queued SLEW_TO_TARGET commands as the pending ACS attitude intent.
- Cancel older pending target slews when ACS accepts a newer target slew or pass-start attitude decision.
- Log each canceled pending slew with its scheduled time, obsid, and superseding command.
- Add regression coverage for science-slew supersession, START_PASS supersession, and non-superseding commands preserving pending slews.

### DIFF
--- a/conops/simulation/acs.py
+++ b/conops/simulation/acs.py
@@ -205,9 +205,9 @@ class ACS:
                 ),
             )
 
-    @classmethod
-    def _command_label(cls, command: ACSCommand) -> str:
-        obsid = cls._command_obsid(command)
+    @staticmethod
+    def _command_label(command: ACSCommand) -> str:
+        obsid = ACS._command_obsid(command)
         if obsid is None:
             return command.command_type.name
         return f"{command.command_type.name} obsid={obsid}"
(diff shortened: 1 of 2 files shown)


conops/simulation/acs.py
```diff
@@ -169,6 +175,49 @@ def enqueue_command(self, command: ACSCommand) -> None:
...
+                    f"SLEW_TO_TARGET command scheduled for "
+                    f"{unixtime2date(canceled.execution_time)} "
+                    f"(obsid={canceled_obsid}) - superseded by "
+                    f"{replacement_label}"
+                ),
+            )
+
+    @classmethod
+    def _command_label(cls, command: ACSCommand) -> str:
+        obsid = cls._command_obsid(command)
```
jak574: As `_command_obsid` is a staticmethod, this should be too.
```suggestion
    @staticmethod
    def _command_label(command: ACSCommand) -> str:
        obsid = ACS._command_obsid(command)
```

jak574 (dismissed):
lgtm, one minor edit requested.

PR #145, opened 2026-05-18
fix: sync plan slew metadata with ACS execution

Tohuvavohu:
Closes #142.

The exported plan was carrying slew metadata from queue-time estimates, while ACS later recomputed the executed slew from the actual spacecraft attitude and roll. That let `slewtime` / `slewdist` in the JSON drift from what ACS actually commanded. A full Tamalpais DITL regeneration also showed that when ACS executed more than one science slew in the same timestep, syncing only from `acs.current_slew` could leave an earlier exported entry with its queue-time `begin`.

This PR makes the ACS-executed `Slew` objects the source of truth for science-plan slew metadata:

- copy `slewtime`, `slewdist`, roll, begin time, and slew path from the scheduled `Slew` onto the selected target
- resync exported plan entries from newly executed ACS science slew commands, so multiple same-timestep slews are handled
- keep the active-slew sync path for direct/current ACS slew state
- include executed slew distance in ACS `SLEW` log events
- add regression coverage for command metadata propagation, exported plan-entry resync, same-timestep executed slew sync, rejection handling, and ACS `SLEW` event parity

Validation:
- `pytest`
- `pytest tests/acs tests/scheduler_queue/test_queue_ditl.py`
- `prek run --all-files`
- Full Tamalpais DITL plan generation for 2026-03-01T18:00:00Z to 2026-03-02T18:00:00Z, comparing serialized JSON science entries against executed ACS slew commands and ACS `SLEW` log events: 0 metadata mismatches, 0 event mismatches

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -700,9 +700,7 @@ class QueueDITL(DITLMixin, DITLStats):
 
     @staticmethod
     def _entry_matches_slew(entry: PlanEntry, slew: Slew) -> bool:
-        return entry.obstype == ObsType.AT and getattr(entry, "obsid", None) == getattr(
-            slew, "obsid", None
-        )
+        return entry.obstype == ObsType.AT and entry.obsid == slew.obsid
 
     def _sync_slew_metadata(self, slew: Slew) -> None:
         if slew.obstype != ObsType.PPT:
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -151,6 +151,7 @@ class QueueDITL(DITLMixin, DITLStats):
         # successful dispatch; None during a pass (spacecraft is occupied).
         self._ppt_unavailable: bool | None = None
         self._planned_gsp_keys: set[tuple[str, float, float]] = set()
+        self._synced_executed_slew_count = 0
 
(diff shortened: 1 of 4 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -684,6 +686,60 @@ def _close_ppt_timeline_if_needed(self, utime: float) -> None:
...
+        entry.slewpath = slew.slewpath
+        entry.roll = float(slew.endroll)
+        if update_end:
+            entry.end = int(slew.slewstart + slew.slewtime + entry.ss_max)
+
+    @staticmethod
+    def _entry_matches_slew(entry: PlanEntry, slew: Slew) -> bool:
+        return entry.obstype == ObsType.AT and getattr(entry, "obsid", None) == getattr(
+            slew, "obsid", None
+        )
```
jak574: No `getattr` here please, both `Slew` and `PlanEntry` initialize `obsid` on instantiation, so there's no need to check if it exists.
```suggestion
    @staticmethod
    def _entry_matches_slew(entry: PlanEntry, slew: Slew) -> bool:
        return entry.obstype == ObsType.AT and entry.obsid == slew.obsid
```

jak574 (dismissed):
lgtm, with one edit.

PR #147, opened 2026-05-19
fix: validate exported plan entries against ACS execution

Tohuvavohu:
## Summary
- fail QueueDITL before returning when exported science/contact plan entries do not match recorded ACS telemetry
- validate science obsid/pointing and GSP contact mode/obsid/pass pointing profile
- propagate GSP obsid through ACS pass telemetry so contact validation compares against executed pass state

This prevents a mismatched plan from being produced as valid

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -608,6 +608,11 @@ class QueueDITL(DITLMixin, DITLStats):
     def _science_start_time(self, entry: PlanEntry) -> float:
         return float(entry.begin) + max(0.0, float(getattr(entry, "slewtime", 0.0)))
 
+    def _window_indices(self, start: float, end: float) -> range:
+        lo = max(0, int(self.ephem.index(dtutcfromtimestamp(start))))
+        hi = int(self.ephem.index(dtutcfromtimestamp(end)))
+        return range(lo, hi)
+
     def _execution_mismatch(
         self,
         utime: float,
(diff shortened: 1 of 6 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -543,6 +562,283 @@ def _is_short_science_entry(self, entry: PlanEntry) -> bool:
...
+        end = float(entry.end)
+        mismatches: list[PlanExecutionMismatch] = []
+        if end <= start:
+            return mismatches
+
+        samples = 0
+        science_samples = 0
+        for i, utime in enumerate(self.utime):
+            if not (start <= utime < end):
+                continue
```
jak574: Maybe a performance hit here rather than just figuring out what the indices of start and end are and running between those. Could add a helper as such:

```python
    def _window_indices(self, start: float, end: float) -> range:
        lo = max(0, self.ephem.index(dtutcfromtimestamp(start)))
        hi = self.ephem.index(dtutcfromtimestamp(end))
        return range(lo, hi)
```
and then 
```suggestion
        for i in self._window_indices(start, end):
            utime = self.utime[i]
```

conops/ditl/queue_ditl.py
```diff
@@ -543,6 +562,283 @@ def _is_short_science_entry(self, entry: PlanEntry) -> bool:
...
+                    "contact",
+                    "pass_profile_missing",
+                    f"obsid {int(entry.obsid)} station {getattr(entry, 'station', None)}",
+                    obsid=int(entry.obsid),
+                )
+            )
+
+        for i, utime in enumerate(self.utime):
+            if not (start <= utime < end):
+                continue
```
jak574: This scans through the entire simulation time every time.
Add helper mentioned in previous comment then change to:
```suggestion
        for i in self._window_indices(start, end):
            utime = self.utime[i]
```

jak574 (changes requested):
One minor change that could address a possible performance issue.

jak574 (approved):
lgtm

PR #153, opened 2026-05-23
fix: stop charge completion from reviving stale targets

### LINKED ISSUE #148: Fix stale charge-return target after emergency charging
The bug is simple: after emergency charging ends, ACS can return to a stale `last_ppt` even though QueueDITL has already selected the next target and recorded it in the plan.

That lets the plan and execution diverge. In the bad Tamalpais run, the plan recorded target `10132`, but ACS actually returned to/executed stale target `10454`.

This should not be possible. PR #147 adds the guardrail so this fails fast, but it does not fix the scheduling/state-machine bug.

Fix the post-charge path so there is one owner for the next target after charging. ACS should not independently enqueue a stale `last_ppt` return when QueueDITL has selected a new observation.

Acceptance criteria:
- regression test reproduces the stale charge-return mismatch
- planned obsid/attitude match executed ACS obsid/attitude after charge exit
- PR #147's plan/execution invariant passes for this case

Tohuvavohu:
Fixes #148.

ACS no longer enqueues a return slew to last_ppt when charging ends. QueueDITL owns the next target after charging, so ACS only clears charging state and waits for the next queued target.

Validation:
- pytest tests/acs/test_acs_coverage.py -k end_battery_charge
- pre-commit run --all-files

### DIFF
--- a/conops/simulation/acs.py
+++ b/conops/simulation/acs.py
@@ -1054,10 +1054,7 @@ class ACS:
             )
 
     def _end_battery_charge(self, utime: float) -> None:
-        """Handle END_BATTERY_CHARGE command execution.
-
-        Terminates charging mode by returning to previous science pointing.
-        """
+        """Handle END_BATTERY_CHARGE command execution."""
         self._log_or_print(utime, "CHARGING", "Ending battery charge")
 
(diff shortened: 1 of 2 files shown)


jak574 (approved):
Yeah, good catch on this one. The previous behavior is wrong.

PR #154, opened 2026-05-23
fix: drop interrupted science entries during charging

Tohuvavohu:
Fixes #152.

When emergency charging interrupts active science, QueueDITL now removes the active science plan entry that matches the interrupted target and keeps short science retryable.

Validation:
- pytest tests/scheduler_queue/test_queue_ditl.py -k charging
- pre-commit run --all-files


jak574 (approved):
lgtm

PR #155, opened 2026-05-23
fix: synchronize GSP plan entries with ACS execution

Tohuvavohu:
Fixes #150.

GSP entries now line up with ACS execution instead of being exported before ACS has actually processed the pass command. QueueDITL re-reads ACS pointing and mode after a due pass command, and pass pre-slew timing uses the same Slew model ACS executes, including roll.

Validation:
- pytest tests/passes/test_passes.py::TestPassTimeToSlew tests/scheduler_queue/test_queue_ditl.py::TestCalcMethod::test_calc_processes_due_pass_command_same_step
- pre-commit run --all-files

### DIFF
--- a/conops/simulation/passes.py
+++ b/conops/simulation/passes.py
@@ -9,6 +9,10 @@ from ..common.enums import AntennaType
 from ..common.vector import radec2vec, rotvec, separation, vec2radec
 from ..config import Constraint, GroundStationRegistry, MissionConfig
 from ..config.constants import DTOR
+from .slew import Slew
+
+# Current ground-pass model tracks RA/Dec while holding a fixed roll.
+GSP_TRACK_ROLL = 0.0
 
 
 def pass_slew_trigger_buffer(step_size: float) -> float:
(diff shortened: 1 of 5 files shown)


conops/simulation/passes.py
```diff
@@ -227,7 +228,31 @@ def calculate_data_volume(self, band: str, direction: str = "downlink") -> float
...
+        if self.config.spacecraft_bus.attitude_control is None:
+            raise ValueError("ACS config must be set to calculate slew time")
+
+        slew = Slew(config=self.config)
+        slew.startra = ra
+        slew.startdec = dec
+        slew.startroll = roll
+        slew.endra = target_ra
+        slew.enddec = target_dec
+        slew.slewstart = utime
```
jak574, line 254: You're setting `startroll` but not `endroll`?

Yeah,  I understand this is a whole thing. How does the roll change when you're tracking a ground station? I don't know. Issue is that in some cases roll changes can change slew time, so it's probably important to know this, on the flip side, it's likely a spacecraft dependent how roll during a groundstation pass is implemented.

jak574 (approved):
Approved with comment for future thought.

conops/simulation/passes.py
Tohuvavohu: Addressed by making the current GSP roll policy explicit instead of relying on Slew.endroll's default. The pass model now has a shared GSP_TRACK_ROLL constant, Pass.time_to_slew uses it for the predicted slew, and QueueDITL sets the same endroll on the queued ACS GSP slew. 

This keeps the current model as fixed-roll ground-pass tracking; a spacecraft-specific roll policy can replace that constant/model later.

PR #156, opened 2026-05-23
fix: use telemetry timestamps for validation windows

### LINKED ISSUE #151: Plan/execution validation uses nearest ephem sample for interval bounds
The plan/execution invariant should check telemetry samples inside the scheduled interval.

Right now _window_indices() uses ephem.index(dtutcfromtimestamp(...)). That can return the nearest ephemeris sample, which pulls samples from just outside the interval into validation. The invariant can then report false mismatches at boundaries.

Use telemetry timestamps directly and half-open interval bounds instead.

Tohuvavohu:
Fixes #151.

Plan/execution validation now indexes telemetry windows from self.utime with half-open interval bounds. That keeps samples just outside the scheduled interval from being pulled into validation by nearest ephemeris lookup.

Validation:
- pytest tests/scheduler_queue/test_queue_ditl.py::TestPlanExecutionValidation
- pre-commit run --all-files

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -1,3 +1,4 @@
+from bisect import bisect_left
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Any, cast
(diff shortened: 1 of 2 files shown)


jak574 (approved):
Even more optimized. lgtm

PR #158, opened 2026-05-23
fix: use ObsType for internal obstype fields

Tohuvavohu:
Fixes #157.

Make internal obstype fields use ObsType instead of raw strings:
- Pass.obstype now defaults to ObsType.GSP
- ACSCommand.obstype now defaults to ObsType.PPT
- tests that asserted the old internal string contract now assert ObsType values

String values are still accepted at serialization/config boundaries through pydantic enum coercion.

Scope note: this PR intentionally does not mass-edit every test mock that assigns string obstypes. Those mocks still work because ObsType is a str enum, and changing all of them would be broad churn. The production/internal model fields are the contract this PR fixes.

Validation:
- pytest tests/passes/test_passes.py::TestPassInitialization tests/acs/test_acs_safe_mode.py::TestSafeModeInitialization tests/acs/test_acs.py::TestACSInitialization::test_acs_initialization tests/acs/test_acs.py::TestACSAttributes::test_obstype_attribute tests/plan_entry/test_plan_entry.py::TestPlanEntryInit::test_init_with_constraint_and_acs tests/pointing/test_pointing.py::TestPointing::test_obstype_at tests/scheduler_queue/test_queue_ditl.py::TestCheckAndManagePasses::test_check_and_manage_passes_slew_to_pass_has_gsp_obstype
- pre-commit run --all-files

### DIFF
--- a/conops/simulation/acs_command.py
+++ b/conops/simulation/acs_command.py
@@ -2,7 +2,7 @@ from typing import TYPE_CHECKING
 
 from pydantic import BaseModel, ConfigDict
 
-from ..common import ACSCommandType
+from ..common import ACSCommandType, ObsType
 
 if TYPE_CHECKING:
     from .slew import Slew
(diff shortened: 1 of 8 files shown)


jak574 (approved):
Necessary cleanup. lgtm

PR #160, opened 2026-05-23
fix: export executed GSP slew metadata

Tohuvavohu:
Fixes #159.

GSP plan entries were exporting the reserved pre-contact window as slewtime and leaving slewdist at 0.0. That made downstream plan stats and visualization show contacts as zero-distance maneuvers and could label pre-contact buffer as slew.

This changes GSP export so:
- the queued GSP entry gets physical slewtime/slewdist from the pre-pass slew command
- after ACS executes and recomputes the maneuver from current attitude, the GSP plan entry is synced to the executed slew start, slewtime, and slewdist
- GSP exposure/contact duration stays tied to the contact window, not to the activity window minus slewtime

The slew distance is the maneuver to the ground-station pass start pointing. It is not the angular distance swept while tracking during the contact.

Validation:
- pytest tests/scheduler_queue/test_queue_ditl.py::TestCheckAndManagePasses tests/scheduler_queue/test_queue_ditl.py::TestFetchNewPPT tests/plan/test_plan_schema.py tests/plan_entry/test_plan_entry.py
- ruff check conops/ditl/queue_ditl.py conops/targets/plan_entry.py conops/targets/plan_schema.py tests/scheduler_queue/test_queue_ditl.py
- pre-commit run --all-files

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -169,6 +169,7 @@ class QueueDITL(DITLMixin, DITLStats):
         # successful dispatch; None during a pass (spacecraft is occupied).
         self._ppt_unavailable: bool | None = None
         self._planned_gsp_keys: set[tuple[str, float, float]] = set()
+        self._gsp_slew_plan_entries: dict[Slew, PlanEntry] = {}
         self._synced_executed_slew_count = 0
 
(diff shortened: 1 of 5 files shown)


jak574 (approved):
lgtm (after some thinking about it)

PR #162, opened 2026-05-24
refactor: clean up plan_schema.py

jak574:
This pull request refactors the way metadata fields are handled in the `AttitudeTimeseriesSchema` and `PlanSchema` classes to improve data consistency and reduce manual bookkeeping. The main change is to compute `num_samples` and `num_entries` dynamically from their respective lists, rather than storing and updating them manually. Additionally, the logic for correcting legacy metadata has been simplified.

Key changes:

**Dynamic Computed Fields:**
* Replaced the stored `num_samples` and `num_entries` integer fields in `AttitudeTimeseriesSchema` and `PlanSchema` with computed properties that automatically return the length of the `samples` and `entries` lists, respectively. This ensures that these counts are always accurate and up to date. [[1]](diffhunk://#diff-30eec4d8aaeb66d37ed45dcaa0a464c60de689b76175daba86445f1fb503749bL173-R180) [[2]](diffhunk://#diff-30eec4d8aaeb66d37ed45dcaa0a464c60de689b76175daba86445f1fb503749bL236-R246)

**Simplified Legacy Metadata Correction:**
* Removed the model validators that previously reconciled metadata fields like `num_samples` and `num_entries` after model creation. The new approach only updates `start` and `end` fields if they are missing (i.e., set to `0.0`), using the first and last entry times, making the logic simpler and more robust for legacy JSON file support. [[1]](diffhunk://#diff-30eec4d8aaeb66d37ed45dcaa0a464c60de689b76175daba86445f1fb503749bL191-L195) [[2]](diffhunk://#diff-30eec4d8aaeb66d37ed45dcaa0a464c60de689b76175daba86445f1fb503749bL288-R298)

**Imports:**
* Added import for `computed_field` to support the new computed property decorators.

### DIFF
--- a/conops/targets/plan_schema.py
+++ b/conops/targets/plan_schema.py
@@ -30,6 +30,7 @@ from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
+    computed_field,
     field_serializer,
     field_validator,
     model_validator,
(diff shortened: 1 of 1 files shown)


Tohuvavohu (approved):
lgtm

PR #164, opened 2026-05-24
fix: make plan scheduling deterministic

Tohuvavohu:
Closes #163.

Identical inputs were not guaranteed to produce identical schedules because target merit ties and ground-station pass probabilities used process-global NumPy randomness.

This PR:

- removes random target merit tie-breaking; equal-merit targets now keep queue order
- adds `MissionConfig.random_seed`
- makes `PassTimes` use a seeded/injected RNG for probabilistic pass selection
- avoids consuming RNG when pass probability is 0 or 1
- adds tests for deterministic ordering and seeded pass requests

Validated:

- `pytest tests/target_queue/test_target_queue.py tests/passes/test_passes.py`

### DIFF
--- a/conops/config/config.py
+++ b/conops/config/config.py
@@ -26,6 +26,10 @@ class MissionConfig(ConfigModel):
     """
 
     name: str = Field(default="Default Config", description="Mission name/identifier")
+    random_seed: int | None = Field(
+        default=None,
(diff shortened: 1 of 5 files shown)


jak574 (approved):
Good idea. LGTM

PR #166, opened 2026-05-24
fix: prevent stale plan entry slew sync

Tohuvavohu:
Closes #165.

A target can be retried with the same obsid after an earlier constrained or under-collected attempt. The exported plan metadata sync was matching science slew metadata by obsid only, so a later retry could rewrite an older closed plan entry with the later slew start while leaving the old end time in place.

This PR:

- only syncs science slew metadata into the active/open entry for that slew
- validates exported plan structure before export so invalid intervals or non-monotonic entries fail inside COAST
- adds regression tests for duplicate-obsid stale sync and structural plan validation

Validated:

- `pytest tests/scheduler_queue/test_queue_ditl.py`

I also validated this fix with Tamalpais full generation locally; JSON, FlightJAS, orbit visualizer payload, timelines, event logs, and manifest all complete when local `fjplan` and `orbit_visualizer` packages are on `PYTHONPATH`.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -676,6 +676,45 @@ class QueueDITL(DITLMixin, DITLStats):
             obsid=int(entry.obsid),
         )
 
+    def _validate_plan_entry_structure(self) -> list[PlanExecutionMismatch]:
+        mismatches: list[PlanExecutionMismatch] = []
+        previous_begin: float | None = None
+        for index, entry in enumerate(self.plan):
+            obstype = self._entry_obstype(entry)
+            if obstype is None or obstype == ObsType.PPT:
+                continue
+            begin = float(entry.begin)
+            end = float(entry.end)
+            obsid = int(entry.obsid) if entry.obsid is not None else None
+            if end < begin:
+                mismatches.append(
+                    self._execution_mismatch(
+                        begin,
+                        "plan",
+                        "invalid_interval",
+                        (
+                            f"entry {index} obsid {obsid} ends before it begins "
+                            f"({end:.0f} < {begin:.0f})"
+                        ),
+                        obsid=obsid,
+                    )
+                )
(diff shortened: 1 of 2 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -676,6 +676,45 @@ def _pointing_mismatch(
...
+        mismatches: list[PlanExecutionMismatch] = []
+        previous_begin: float | None = None
+        for index, entry in enumerate(self.plan):
+            obstype = self._entry_obstype(entry)
+            if obstype is None or obstype == ObsType.PPT:
+                continue
+            begin = float(entry.begin)
+            end = float(entry.end)
+            obsid = int(entry.obsid) if entry.obsid is not None else None
+            if end < begin:
```
jak574: This catches cases where `end < begin`, but what about cases where `end == begin`? Those I would think are equally invalid. Basically what I'm saying is:

```suggestion
            if end <= begin:
```

conops/ditl/queue_ditl.py
```diff
@@ -1004,7 +1046,15 @@ def _apply_slew_metadata(
...
     @staticmethod
     def _entry_matches_science_slew(entry: PlanEntry, slew: Slew) -> bool:
-        return entry.obstype == ObsType.AT and entry.obsid == slew.obsid
+        if entry.obstype != ObsType.AT or entry.obsid != slew.obsid:
+            return False
+
+        entry_begin = float(entry.begin)
+        entry_end = float(entry.end)
+        slew_start = float(slew.slewstart)
+        has_placeholder_end = entry_end >= entry_begin + 86400
```
jak574: Avoid magic numbers, where a constant exists.
```suggestion
        has_placeholder_end = entry_end >= entry_begin + DAY_SECONDS
```

jak574 (changes requested):
Minor changes, other lgtm.

jak574 (approved):

PR #169, opened 2026-05-25
fix: validate executed science/contact activity before export

### LINKED ISSUE #167: Closed observations can continue as SCIENCE outside the exported plan
COAST can close a science plan entry and clear the active PPT, but ACS keeps holding the last slew attitude and get_mode() can still report SCIENCE.

That means telemetry can keep saying SCIENCE obsid=<old target> after the exported JSON entry ended. If that held attitude later goes into Earth/Sun/etc. constraint, the generated products look like the plan violates constraints, but the bad samples are not represented by the plan entry.

The deeper problem is that validation is mostly plan-entry -> execution. We also need execution -> plan.

Expected:
- once a scheduled observation is closed, holding that attitude is not reported as SCIENCE
- every executed SCIENCE sample is covered by a matching exported AT/TOO entry
- every executed PASS sample is covered by a matching exported GSP entry
- every exported science collection sample is checked against the science constraints
- GSP is not checked against science Earth-limb keepout, because contacts intentionally point at Earth

This should fail generation before writing a JSON plan that does not match the executed ACS timeline.

Tohuvavohu:
## Summary
- reject generated plans when executed SCIENCE samples violate the applicable science constraints
- require executed SCIENCE/PASS telemetry to be covered by the exported AT/TOO/GSP entries so activity telemetry cannot drift away from the plan
- make ACS report held post-activity attitudes as IDLE instead of continuing to report stale SCIENCE obsids after the activity is closed

This PR fixes the immediate bug where a plan can look valid while executed science/contact telemetry does not match the exported activity windows. It validates activity-scoped executed telemetry before attitude export. It does not yet run every attitude sample in every ACS mode through a full mode-specific constraint policy; that is follow-up work.

Closes #167.

## Validation
- `pytest`
- `prek run --all-files`
- Tamalpais integration with #164 + #166 + this branch now fails before export on #168, which is the intended behavior until dropped under-collected retries are fixed.

### DIFF
--- a/tests/scheduler_queue/test_queue_ditl.py
+++ b/tests/scheduler_queue/test_queue_ditl.py
@@ -20,6 +20,7 @@ from conops import (
 from conops.common.enums import ObsType
 from conops.config.config import MissionConfig
 from conops.ditl.telemetry import Housekeeping
+from conops.simulation.acs import IDLE_OBSID
 from conops.simulation.passes import GSP_TRACK_ROLL
 from conops.targets import PlanEntry, PlanSchema
 
@@ -313,7 +314,21 @@ class TestDetermineMode:
         acs.saa = None
 
         mode = acs.get_mode(1000.0)
+        assert mode == ACSMode.IDLE
+
+        science_slew = Slew(config=mock_config)
+        science_slew.obstype = ObsType.PPT
+        science_slew.obsid = 42
+        science_slew.slewstart = 0.0
+        science_slew.slewend = 0.0
+        acs.last_slew = science_slew
+        acs.science_observation_active = True
+        mode = acs.get_mode(2000.0)
         assert mode == ACSMode.SCIENCE
+
+        acs.end_science_observation()
+        mode = acs.get_mode(1000.0)
+        assert mode == ACSMode.IDLE
 
 
 class TestHandlePassMode:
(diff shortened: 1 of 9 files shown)


tests/scheduler_queue/test_queue_ditl.py
```diff
@@ -1430,14 +1445,103 @@ def test_validation_uses_samples_inside_scheduled_interval(
...
+        science_slew.obsid = entry.obsid
+        science_slew.slewstart = 1000.0
+        science_slew.slewend = 1100.0
+        science_slew.slewtime = 100.0
+        science_slew.startra = 0.0
+        science_slew.startdec = 0.0
+        science_slew.endra = entry.ra
+        science_slew.enddec = entry.dec
+        real_acs.last_slew = science_slew
+        real_acs.science_observation_active = True
```
jak574, line 1526: It's cleaner to group asserts after assignments.
```suggestion
        real_acs.science_observation_active = True
        real_acs._check_constraints = Mock()
        real_acs._check_star_tracker_constraints = Mock()
        real_acs._check_radiator_constraints = Mock()
```

tests/scheduler_queue/test_queue_ditl.py
```diff
@@ -1430,14 +1445,103 @@ def test_validation_uses_samples_inside_scheduled_interval(
...
+
+        queue_ditl._terminate_ppt(1300.0, reason="Target constrained")
+
+        assert queue_ditl.ppt is None
+        assert real_acs.science_observation_active is False
+        assert real_acs.get_mode(1300.0) == ACSMode.IDLE
+        real_acs._check_constraints = Mock()
+        real_acs._check_star_tracker_constraints = Mock()
+        real_acs._check_radiator_constraints = Mock()
```
jak574:
```suggestion
```

jak574 (changes requested):

Tohuvavohu:
@jak574 fixed

jak574 (approved):

PR #171, opened 2026-05-25
fix: validate full attitude constraint timeline

Tohuvavohu:
## Summary
- add a pre-export validation pass over the complete executed attitude telemetry timeline
- apply the full mission constraint set in SCIENCE and CHARGING modes
- apply hard keepout constraints in PASS, SLEWING, IDLE, SAFE, and SAA so ground contacts do not inherit science Earth-limb rules
- report timestamp, mode, obsid, RA/Dec, roll, policy, and violated constraint when validation fails

The actual constraint geometry and thresholds remain config-driven. This PR only defines which constraint policy is applied to each ACS mode during export validation.

Stacked on #169. Closes #170.

## Validation
- `pytest`
- `prek run --all-files`

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -18,7 +18,7 @@ from ..common import (
 )
 from ..common.enums import ACSCommandType
 from ..common.vector import attitude_to_quat
-from ..config import DAY_SECONDS, MissionConfig
+from ..config import DAY_SECONDS, AttitudeConstraintPolicy, MissionConfig
 from ..simulation.acs_command import ACSCommand
 from ..simulation.emergency_charging import EmergencyCharging
 from ..simulation.passes import GSP_TRACK_ROLL, Pass, pass_slew_trigger_buffer
@@ -602,6 +602,7 @@ class QueueDITL(DITLMixin, DITLStats):
             "utime": len(self.utime),
             "ra": len(self.ra),
             "dec": len(self.dec),
+            "roll": len(self.roll),
             "obsid": len(self.obsid),
             "mode": len(self.mode),
         }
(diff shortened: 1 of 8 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -851,6 +846,66 @@ def _validate_execution_is_planned(self) -> list[PlanExecutionMismatch]:
...
+    def _attitude_constraint_name_for_sample(
+        self, index: int, mode: ACSMode
+    ) -> tuple[str, str] | None:
+        """Return the violated constraint name and policy for one attitude sample."""
+        ra = float(self.ra[index])
+        dec = float(self.dec[index])
+        roll = float(self.roll[index])
+        utime = self.utime[index]
+        policy = self.config.attitude_constraint_policy_for_mode(mode)
+        if policy == AttitudeConstraintPolicy.FULL_MISSION:
```
jak574, line 878: I think it would be good to be clear that if the policy is NONE, we just exit, this is implied, but can be made implicitly by adding these lines at the top of the `if`s:

```python
        if policy == AttitudeConstraintPolicy.NONE:
            return None
```
That way, logic things doesn't get confusing if more complexity is added later on.

conops/ditl/queue_ditl.py
```diff
@@ -676,15 +677,32 @@ def _pointing_mismatch(
...
+        )
+
+    def _unknown_mode_mismatch(self, index: int) -> PlanExecutionMismatch:
+        return self._execution_mismatch(
+            self.utime[index],
+            "attitude",
+            "unknown_mode",
+            f"cannot apply constraint policy for mode {self.mode[index]}",
+            obsid=int(self.obsid[index]),
         )
```
jak574, line 706: Should add test coverage for this.

jak574 (changes requested):
A few changes requested.

jak574 (approved):
lgtm

PR #175, opened 2026-05-25
fix: retry equal-merit target selection

Tohuvavohu:
## Summary
- add a seeded deterministic tie-break for equal-merit targets so ordering does not collapse to checked-in catalog order
- retry target selection in the same scheduler tick when downstream roll validation rejects the selected PPT
- cover equal-merit ordering and locked-roll retry behavior in tests

Fixes #174.

This is stacked on #164 because the regression is introduced by the deterministic scheduling change; the PR should be retargeted to main after #164 lands, or merged after it.

The repeated-obsid stale plan entry sync exposed during full Tamalpais export is intentionally left to #166.

## Validation
- pytest tests/target_queue/test_target_queue.py tests/scheduler_queue/test_queue_ditl.py -q
- pytest -q
- prek run --all-files
- representative Tamalpais in-memory run before removing the #166-overlap commit: 90 entries (72 AT, 15 CHARGE, 3 GSP), 187 locked-roll skips; full export still requires #166 for repeated-obsid plan-entry sync

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -1805,15 +1805,11 @@ class QueueDITL(DITLMixin, DITLStats):
                         target_dec=self.ppt.dec,
                     )
                     if not _valid_ranges:
-                        obsid = self.ppt.obsid
                         self.log.log_event(
                             utime=utime,
                             event_type="QUEUE",
-                            description=(
-                                f"Target {obsid} skipped - no valid roll satisfies "
-                                "all constraints at this time"
-                            ),
-                            obsid=obsid,
+                            description=f"Target {self.ppt.obsid} skipped — no valid roll satisfies all constraints at this time",
+                            obsid=self.ppt.obsid,
                             acs_mode=self.acs.acsmode,
                         )
                         self._retry_fetch_without_current_ppt(utime, ra, dec)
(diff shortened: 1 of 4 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -1805,15 +1805,18 @@ def _fetch_new_ppt(self, utime: float, ra: float, dec: float) -> None:
...
                         self.log.log_event(
                             utime=utime,
                             event_type="QUEUE",
-                            description=f"Target {self.ppt.obsid} skipped — no valid roll satisfies all constraints at this time",
-                            obsid=self.ppt.obsid,
+                            description=(
+                                f"Target {obsid} skipped - no valid roll satisfies "
+                                "all constraints at this time"
+                            ),
+                            obsid=obsid,
```
jak574: Don't really understand why this change is needed?

conops/ditl/queue_ditl.py
```diff
@@ -1852,19 +1855,19 @@ def _fetch_new_ppt(self, utime: float, ra: float, dec: float) -> None:
...
                         utime=utime,
                         event_type="QUEUE",
                         description=(
-                            f"Target {self.ppt.obsid} skipped — locked roll "
+                            f"Target {obsid} skipped - locked roll "
                             f"{slew.endroll:.1f}° violates constraint "
                             f"within minimum observation window (at t+{t_unix - obs_start_time:.0f}s)"
                         ),
-                        obsid=self.ppt.obsid,
+                        obsid=obsid,
```
jak574: Why the need to make `obsid` here instead of just using `self.ppt.obsid`. Is this some weird Python reference thing?

jak574 (changes requested):
FYI I rebased this, so I could see what's new. There are some apparently somewhat arbitrary changes. Can you explain?

jak574 (approved):

PR #177, opened 2026-05-26
fix: track GSP contacts with configured antenna vector

Tohuvavohu:
## Summary
- add an explicit fixed antenna body-frame boresight vector, defaulting to legacy -X
- generate GSP attitude profiles so a FIXED configured antenna vector tracks the station line of sight
- carry GSP roll through ACS execution, plan export, and plan/execution validation
- keep the old azimuth/elevation helper as legacy compatibility metadata, with a deprecation warning for nonzero legacy az/el config
- avoid silently treating OMNI/GIMBALED antennas as legacy fixed -X body tracking

Closes #172.
Closes #173.

Follow-up: #181 tracks proper OMNI/GIMBALED ground-contact behavior without body-frame attitude tracking.

## Validation
- `pytest -q`
- `pytest tests/config/test_communications.py tests/passes/test_passes.py tests/passes/test_antenna_offset.py tests/passes/test_passes_communications.py`
- `PREK_HOME=.prek-cache PREK_LOG=.prek-cache/prek.log PATH="coastvenv/bin:$PATH" prek run --all-files`
- representative Tamalpais DITL with fixed +Y antenna override: 3 GSP entries, 24 contact samples, 0 GSP validation mismatches, max +Y-to-station error 8.5e-7 deg, legacy -X-to-station error ~90 deg

### DIFF
--- a/conops/common/vector.py
+++ b/conops/common/vector.py
@@ -194,6 +194,140 @@ def normal_to_euler_deg(
     return 0.0, pitch_deg, yaw_deg
 
 
+def _unit_vector(
+    vector: tuple[float, float, float] | npt.NDArray[np.float64], name: str
+) -> npt.NDArray[np.float64]:
+    arr = np.asarray(vector, dtype=np.float64)
+    norm = float(np.linalg.norm(arr))
+    if norm < 1e-12:
+        raise ValueError(f"{name} must be non-zero")
+    return arr / norm
+
+
+def _perpendicular_reference(
+    axis: npt.NDArray[np.float64],
+    preferred: tuple[float, float, float] | npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    for candidate in (
+        preferred,
+        (0.0, 0.0, 1.0),
+        (0.0, 1.0, 0.0),
+        (1.0, 0.0, 0.0),
+    ):
+        ref = _unit_vector(candidate, "reference vector")
+        projected = ref - float(np.dot(ref, axis)) * axis
+        norm = float(np.linalg.norm(projected))
+        if norm >= 1e-12:
+            return projected / norm
+    raise ValueError("could not find a reference vector perpendicular to axis")
+
+
+def attitude_from_body_axes(
+    body_x_eci: tuple[float, float, float] | npt.NDArray[np.float64],
+    body_z_eci: tuple[float, float, float] | npt.NDArray[np.float64],
+) -> tuple[float, float, float]:
+    """Convert body +X/+Z inertial axes to RA, Dec, and roll in degrees."""
+
+    x_axis = _unit_vector(body_x_eci, "body_x_eci")
+    z_axis_raw = _unit_vector(body_z_eci, "body_z_eci")
+    z_axis = z_axis_raw - float(np.dot(z_axis_raw, x_axis)) * x_axis
+    z_axis = _unit_vector(z_axis, "body_z_eci projected perpendicular to body_x_eci")
+
+    ra_rad, dec_rad = vec2radec(x_axis)
+
+    north = np.array([0.0, 0.0, 1.0], dtype=np.float64)
+    n_proj = north - float(np.dot(north, x_axis)) * x_axis
+    n_norm = float(np.linalg.norm(n_proj))
+    if n_norm < 1e-10:
+        north = np.array([1.0, 0.0, 0.0], dtype=np.float64)
+        n_proj = north - float(np.dot(north, x_axis)) * x_axis
+        n_norm = float(np.linalg.norm(n_proj))
+    n_hat = n_proj / n_norm
+    y_hat = np.cross(n_hat, x_axis)
+    roll_rad = float(np.arctan2(np.dot(z_axis, y_hat), np.dot(z_axis, n_hat)))
+    roll_deg = float(np.rad2deg(roll_rad) % 360.0)
+    if abs(roll_deg - 360.0) < 1e-9:
+        roll_deg = 0.0
+
+    return (
+        float(np.rad2deg(ra_rad)),
+        float(np.rad2deg(dec_rad)),
+        roll_deg,
+    )
+
+
+def body_vector_to_eci(
+    ra_deg: float,
+    dec_deg: float,
+    roll_deg: float,
+    body_vector: tuple[float, float, float] | npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    """Return a body-frame vector expressed in ECI for an RA/Dec/roll attitude.
+
+    This is the inverse of the ``scbodyvector`` body-frame transform.
+    """
+
+    x_hat = radec2vec(np.deg2rad(ra_deg), np.deg2rad(dec_deg))
+    ref = np.array([0.0, 0.0, 1.0], dtype=np.float64)
+    y0 = np.cross(ref, x_hat)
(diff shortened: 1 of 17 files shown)


conops/common/vector.py
```diff
@@ -194,6 +194,140 @@ def normal_to_euler_deg(
...
 
 
+def _unit_vector(
+    vector: tuple[float, float, float] | npt.NDArray[np.float64], name: str
+) -> npt.NDArray[np.float64]:
+    arr = np.asarray(vector, dtype=np.float64)
+    norm = float(np.linalg.norm(arr))
+    if norm < 1e-12:
+        raise ValueError(f"{name} must be non-zero")
+    return arr / norm
```
jak574: So this basically does the same as `vecnorm`, except instead of returning the vector if it's close to zero, it raises an exception. My issue with this is that to a user, raising an exception is a crash. I've not traced through exactly how this happens, but if I'm running a 1000 Monte-Carlo, and the code raises an exception 999 runs in, I'm going to be pissed off.

conops/common/vector.py
```diff
@@ -194,6 +194,140 @@ def normal_to_euler_deg(
...
+        (0.0, 0.0, 1.0),
+        (0.0, 1.0, 0.0),
+        (1.0, 0.0, 0.0),
+    ):
+        ref = _unit_vector(candidate, "reference vector")
+        projected = ref - float(np.dot(ref, axis)) * axis
+        norm = float(np.linalg.norm(projected))
+        if norm >= 1e-12:
+            return projected / norm
+    raise ValueError("could not find a reference vector perpendicular to axis")
```
jak574: Same objection applies here, an obscure exception is raised, but not actually handled in any way.

conops/common/vector.py
```diff
@@ -194,6 +194,140 @@ def normal_to_euler_deg(
...
+def attitude_from_body_axes(
+    body_x_eci: tuple[float, float, float] | npt.NDArray[np.float64],
+    body_z_eci: tuple[float, float, float] | npt.NDArray[np.float64],
+) -> tuple[float, float, float]:
+    """Convert body +X/+Z inertial axes to RA, Dec, and roll in degrees."""
+
+    x_axis = _unit_vector(body_x_eci, "body_x_eci")
+    z_axis_raw = _unit_vector(body_z_eci, "body_z_eci")
+    z_axis = z_axis_raw - float(np.dot(z_axis_raw, x_axis)) * x_axis
+    z_axis = _unit_vector(z_axis, "body_z_eci projected perpendicular to body_x_eci")
```
jak574: Might be missing something here, but can't you just do this here:
```suggestion
    z_axis = _perpendicular_reference(x_axis, z_axis_raw)
```
...might as well use that helper some more!

conops/simulation/passes.py
```diff
@@ -96,26 +110,56 @@ def time_to_pass(self) -> str:
...
         RA/Dec in the pass."""
 
-        # Find the closest time index
-        idx = np.searchsorted(self.utime, utime)
-        # searchsorted will return length if utime is beyond the end
-        if idx >= len(self.utime):
-            return self.ra[-1], self.dec[-1]
+        idx = self._profile_index(utime)
+        if idx is None:
+            return self.gsstartra, self.gsstartdec
```
jak574: Is returning these values here the correct thing to do, or should we return `None, None` instead?

conops/common/vector.py
```diff
@@ -194,6 +194,140 @@ def normal_to_euler_deg(
...
+    axis: npt.NDArray[np.float64],
+    preferred: tuple[float, float, float] | npt.NDArray[np.float64],
+) -> npt.NDArray[np.float64]:
+    for candidate in (
+        preferred,
+        (0.0, 0.0, 1.0),
+        (0.0, 1.0, 0.0),
+        (1.0, 0.0, 0.0),
+    ):
+        ref = _unit_vector(candidate, "reference vector")
```
jak574: If this `_unit_vector` raises an exception, then it should be trapped here, rather than just causing a crash. Maybe just skip to the next candidate?
```suggestion
        try:
            ref = _unit_vector(candidate, "reference vector")                                                                                  
        except ValueError:                                                                                            
            continue
```

jak574 (changes requested):

conops/config/communications.py
```diff
@@ -82,13 +89,17 @@ class AntennaPointing(ConfigModel):
         le=180.0,
     )
```
jak574, line 92: As `fixed_azimuth_deg` and `fixed_elevation_deg` are deprecated, add a warning:

```python
    @model_validator(mode="after")
    def warn_legacy_azimuth_elevation(self) -> "AntennaPointing":
        if self.fixed_azimuth_deg != 0.0 or self.fixed_elevation_deg != 0.0:
            warnings.warn(
                "AntennaPointing.fixed_azimuth_deg and fixed_elevation_deg are legacy "
                "metadata fields and no longer affect GSP attitude generation. "
                "Set fixed_boresight_body explicitly to the desired body-frame unit vector.",
                DeprecationWarning,
                stacklevel=2,
            )
        return self
```
Or we could remove them all together. If you do that, change the PR title from a "fix" to a "feat" as this is a breaking API change.

A third option would be continue to support their use, but this would involve creating a helper that converts the alt/az angle to a body frame unit vector.

conops/simulation/passes.py
```diff
@@ -391,6 +467,16 @@ def _pass_selection_key(self, gspass: Pass) -> tuple[float, float, float]:
         rate = self._station_downlink_rate_mbps(gspass)
         return (rate * gspass.length, gspass.length, -gspass.begin)
 
+    def _gsp_antenna_boresight_body(self) -> tuple[float, float, float]:
+        communications = getattr(self.config.spacecraft_bus, "communications", None)
+        if communications is None:
+            return LEGACY_GSP_ANTENNA_BODY_VECTOR
+        antenna = communications.antenna_pointing
+        if antenna.antenna_type != AntennaType.FIXED:
+            return LEGACY_GSP_ANTENNA_BODY_VECTOR
```
jak574: IMHO if antenna type != FIXED, then it's GIMBLED or OMNI, neither of which should have a body vector associated with them. So this should return None.
```suggestion
        if antenna.antenna_type != AntennaType.FIXED:
            return None
```
However if you make this change, you will now have to handle this correctly down below, so that gimbled and omni antennas are correctly handled. 

FYI In future I want to add a feature where the Gimbled antenna has a defined range of motion and has the ability to change pointing, but only in certain ACS modes. So Gimbled antenna handling will get more complex in the future, for now it's just assumed that it's the same as Omni and a stub.

conops/simulation/passes.py
```diff
@@ -524,60 +600,49 @@ def get(self, year: int, day: int, length: int = 1) -> None:
                     )
                     if should_schedule:
                         # Schedule this pass if the dice roll allows it
+                        target_vectors = -gs_to_sat_unit[start_idx:end_idx]
+                        target_ra, target_dec = np.degrees(vec2radec(target_vectors.T))
+                        antenna_boresight = self._gsp_antenna_boresight_body()
+                        attitude_profile: list[tuple[float, float, float]] = []
```
jak574, line 608: Add 
```
                        if antenna_boresight is None:
                            continue
```
here if the change above is made.

jak574 (changes requested):

jak574 (approved):
lgtm

PR #179, opened 2026-05-26
fix: keep executed science attempts in exported plan

Tohuvavohu:
## Summary
- End ACS science collection immediately when emergency charging interrupts an observation.
- Stop science-mode bookkeeping for the timestep once a charging PPT is created.
- Keep science plan entries that actually collected data, even when they are shorter than `ss_min`; only zero-collection science entries are dropped.

## Why
The attitude timeseries can contain real SCIENCE samples for partial observations. Dropping those entries from the exported plan made validation report unplanned science, so the plan no longer represented what ACS executed.


jak574 (approved):
lgtm

PR #182, opened 2026-05-28
fix: clear self.ppt when terminating an immediately-constrained charging PPT

jak574:
## Description

When a charging PPT was created and its target was immediately constrained, `_terminate_charging_ppt` set the PPT's `end == begin` but left `self.ppt` pointing at the orphaned entry. On the next simulation step, `_track_ppt_in_timeline` appended the zero-duration entry to the plan, causing `PlanExecutionMismatchError` with `invalid_interval` mismatches for `999xxx` emergency-charging obsids.

This PR fixes that issue.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -2274,6 +2274,8 @@ class QueueDITL(DITLMixin, DITLStats):
                 self._close_last_plan_entry(utime)
             self.charging_ppt.end = utime
             self.charging_ppt.done = True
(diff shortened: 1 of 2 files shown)


Tohuvavohu (commented):
lgtm, one Q:
should this have a regression test?

that sets `queue_ditl.ppt is queue_ditl.charging_ppt`, calls `_terminate_charging_ppt()`, and verifies `ppt` is also cleared? Ideally also call `_track_ppt_in_timeline()` after termination and verify the zero-duration charging entry is not appended.

PR #184, opened 2026-05-29
fix: keep QueueDITL target state aligned after command churn

Tohuvavohu:
## Summary
- clear a newly selected science PPT when same-tick ACS processing shows its target slew was canceled before execution
- retry rejected PPT candidates iteratively instead of recursively, restoring temporary `done` flags after selection
- close an active PPT at the simulation end boundary rather than at the last recorded sample
- keep this PR scoped to the remaining work not covered by #192

## Why
#192 handles the broad plan/execution bookkeeping drift and dropped-window validation. This follow-up keeps the QueueDITL target pointer aligned with ACS after command churn so a canceled science target cannot become a phantom plan entry, and it removes the recursive retry path that can overflow when many queued targets are rejected in sequence.

This remains stacked on #187 because it uses #187 same-tick ACS command processing.

## Validation
- `/home/aaron/Projects/Cosmic/ops/coast-sim/coastvenv/bin/python -m pytest tests/scheduler_queue/test_queue_ditl.py -q` -> 223 passed
- temporary merge with #192 plus `/home/aaron/Projects/Cosmic/ops/coast-sim/coastvenv/bin/python -m pytest tests/scheduler_queue/test_queue_ditl.py tests/acs/test_acs_coverage.py -q` -> 305 passed

## Relationship to #192
#192 replaces the previous under-collected science / dropped-window validation parts of this branch. Those changes have been removed from this PR.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -2045,27 +2045,25 @@ class QueueDITL(DITLMixin, DITLStats):
         self._retry_fetch_without_current_ppt(utime, ra, dec)
         return True
 
-    def _fetch_new_ppt(
-        self, utime: float, ra: float, dec: float, *, _fetch_loop_active: bool = False
-    ) -> None:
+    def _fetch_new_ppt(self, utime: float, ra: float, dec: float) -> None:
         """Fetch a new pointing target from the queue and enqueue slew command."""
-        if not _fetch_loop_active:
-            self._temporary_rejected_ppts = []
-            self._retry_ppt_fetch_requested = False
-            try:
-                while True:
-                    self._retry_ppt_fetch_requested = False
-                    self._fetch_new_ppt(utime, ra, dec, _fetch_loop_active=True)
-                    if self._retry_ppt_fetch_requested:
-                        continue
-                    return
-            finally:
-                temporary_rejections = self._temporary_rejected_ppts or []
-                for rejected_ppt, was_done in temporary_rejections:
-                    rejected_ppt.done = was_done
-                self._temporary_rejected_ppts = None
+        self._temporary_rejected_ppts = []
+        self._retry_ppt_fetch_requested = False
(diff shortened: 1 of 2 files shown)


conops/ditl/queue_ditl.py
```diff
@@ -2000,8 +2045,27 @@ def _reject_current_ppt_if_insufficient_collect_time(
...
+                    if self._retry_ppt_fetch_requested:
+                        continue
+                    return
+            finally:
+                temporary_rejections = self._temporary_rejected_ppts or []
+                for rejected_ppt, was_done in temporary_rejections:
+                    rejected_ppt.done = was_done
+                self._temporary_rejected_ppts = None
+                self._retry_ppt_fetch_requested = False
+
```
jak574: This is kind of a weird function, as it's calling itself, and relying on passing an argument to determine if it's being called recursively.  I would suggest splitting it into two functions, which makes things less strange.
```suggestion
    def _fetch_new_ppt(self, utime: float, ra: float, dec: float) -> None:
        """Fetch a new pointing target from the queue and enqueue slew command."""
        self._temporary_rejected_ppts = []
        self._retry_ppt_fetch_requested = False
        try:
            while True:
                self._retry_ppt_fetch_requested = False
                self._fetch_new_ppt_inner(utime, ra, dec)
                if self._retry_ppt_fetch_requested:
                    continue
                return
        finally:
            temporary_rejections = self._temporary_rejected_ppts or []
            for rejected_ppt, was_done in temporary_rejections:
                rejected_ppt.done = was_done
            self._temporary_rejected_ppts = None
            self._retry_ppt_fetch_requested = False

    def _fetch_new_ppt_inner(self, utime: float, ra: float, dec: float) -> None:
```

jak574 (changes requested):

jak574 (approved):

PR #188, opened 2026-05-30
fix: only guard science dispatch below battery floor

Tohuvavohu:
## Summary
- Keep QueueDITL fetching/dispatching science while the battery is below the recharge target but still above the configured minimum charge floor.
- Only defer new science dispatch when the battery is below the depth-of-discharge floor (`below_minimum_charge_level`).
- Treat the next pending battery-charge opportunity as a science collection deadline, so COAST does not start targets that cannot meet `ss_min` before charge will preempt them.
- Add regression coverage for both direct PPT fetch and science-completion paths, including pending-charge `ss_min` screening.

## Validation
- `pytest tests/scheduler_queue/test_queue_ditl.py -q` on PR #188 branch: 224 passed.
- Full local stack (#187 + #184 + revised #188 + #189) queue tests: 235 passed.
- Representative Tamalpais plan generation on full local stack: 95 plan entries, 1440 housekeeping samples, 0 plan/execution mismatches.
- Representative event log has 0 dropped under-collected science entries, 0 `battery alert active` deferrals, and 0 `below minimum charge level` deferrals.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -2064,6 +2064,16 @@ class QueueDITL(DITLMixin, DITLStats):
             self._retry_ppt_fetch_requested = False
 
     def _fetch_new_ppt_inner(self, utime: float, ra: float, dec: float) -> None:
+        if self.battery.battery_alert:
+            self.log.log_event(
+                utime=utime,
+                event_type="QUEUE",
+                description="Deferring PPT fetch - battery alert active",
+                acs_mode=self.acs.acsmode,
+            )
+            self._ppt_unavailable = None
+            return
+
         # Don't issue science slews during an active pass - this prevents the
         # teleportation bug where a slew is issued with start position from the
         # pass tracking, but the spacecraft continues following the pass instead.
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -1981,7 +1981,7 @@ class QueueDITL(DITLMixin, DITLStats):
 
         constraint_time = self._ephem_utime_at_or_after(current_time)
         if not self.constraint.in_eclipse(ra=0, dec=0, time=constraint_time):
-            return current_time
+            return constraint_time
 
         for timestamp in self.ephem.timestamp:
             utime = self._ephem_timestamp_to_utime(timestamp)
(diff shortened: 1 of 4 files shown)


Tohuvavohu:
Safety note for review: the ephemeris-grid snap in this PR does not prove continuous-time constraint safety between ephemeris samples. It prevents QueueDITL from calling eclipse constraints at non-ephemeris timestamps by checking the next available ephemeris-grid time, then uses the existing sampled-grid eclipse behavior to decide when a pending charge opportunity can preempt science. Continuous safety still depends on ephemeris step size, margins, and any interval-sweeping validation we add separately.

conops/ditl/queue_ditl.py
```diff
@@ -1828,7 +1828,39 @@ def _next_pass_science_deadline(self, slew_end: float) -> float | None:
...
+        return timestamps[index]
+
+    def _next_charge_science_deadline(self, current_time: float) -> float | None:
+        """Return when pending battery recharge can next preempt science."""
+        if not self.battery.battery_alert or self.charging_ppt is not None:
+            return None
+
+        constraint_time = self._ephem_utime_at_or_after(current_time)
+        if not self.constraint.in_eclipse(ra=0, dec=0, time=constraint_time):
+            return current_time
```
jak574: Should this return `constraint_time`?

conops/ditl/queue_ditl.py
```diff
@@ -1828,7 +1828,39 @@ def _next_pass_science_deadline(self, slew_end: float) -> float | None:
...
+        for timestamp in self.ephem.timestamp:
+            utime = self._ephem_timestamp_to_utime(timestamp)
+            if utime <= current_time:
+                continue
+            if not self.constraint.in_eclipse(ra=0, dec=0, time=utime):
+                return utime
+        return None
+
+    def _next_science_deadline(
+        self, slew_end: float, current_time: float | None = None
```
jak574: Why not make `current_time` required here? Seems like going to `slew_end` could
be problematic.

conops/ditl/queue_ditl.py
```diff
@@ -1844,6 +1876,12 @@ def _next_science_deadline(self, slew_end: float) -> tuple[float, str]:
         if pass_deadline is not None:
             deadlines.append((pass_deadline, "pass"))
 
+        charge_deadline = self._next_charge_science_deadline(
+            current_time if current_time is not None else slew_end
+        )
```
jak574: In this case we just pass `current_time` here.

jak574 (changes requested):
A few questions.

jak574 (approved):

PR #191, opened 2026-05-31
fix: enforce constraint-safe idle holds

Tohuvavohu:
## Summary
- enforce the configured IDLE attitude constraint policy at runtime instead of treating IDLE waits as constraint-free gaps
- replace constrained IDLE holds with a deterministic policy-safe hold attitude before telemetry is recorded
- add an explicit ObsType.IDLE hold so normal idle remediation is not represented as safe mode
- re-query QueueDITL attitude after mode-changing operations so stale science pointings are not recorded as IDLE

## Notes
- This PR does not change the default IDLE policy; it remains hard_keepout unless mission config overrides it.
- Missions that want science-quality idle waits should set IDLE: full_mission in their configuration.

## Tests
- /home/aaron/Projects/Cosmic/ops/coast-sim/coastvenv/bin/python -m pytest tests/acs tests/scheduler_queue/test_queue_ditl.py tests/config/test_config.py -q
- prek run --all-files (passes available hooks; local sphinx-build executable is missing)

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -1399,10 +1399,9 @@ class QueueDITL(DITLMixin, DITLStats):
         return ra, dec, roll, obsid, self.acs.get_mode(utime)
 
     def _has_due_acs_command(self, utime: float) -> bool:
-        command_queue = getattr(self.acs, "command_queue", None)
-        if not isinstance(command_queue, list):
-            return False
-        return any(command.execution_time <= utime for command in command_queue)
+        return any(
+            command.execution_time <= utime for command in self.acs.command_queue
+        )
 
     def _handle_science_mode(
         self, utime: float, ra: float, dec: float, mode: ACSMode
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -458,9 +458,12 @@ class QueueDITL(DITLMixin, DITLStats):
             # Close PPT timeline segment if no active observation
             self._close_ppt_timeline_if_needed(utime)
 
-            # Re-query mode after operations to capture any newly enqueued commands
-            # This prevents 1-step mode gaps when commands are enqueued during this step
-            mode = self.acs.get_mode(utime)
+            # Re-query attitude after operations when they changed mode or queued
+            # an immediate command. This keeps IDLE telemetry from inheriting the
+            # stale science attitude after an observation terminates.
+            ra, dec, roll, obsid, mode = self._refresh_pointing_after_operations(
+                mode, utime, ra, dec, roll, obsid
+            )
 
             # Record pointing and mode
             self._record_pointing_data(ra, dec, roll, obsid, mode)
@@ -1377,6 +1380,29 @@ class QueueDITL(DITLMixin, DITLStats):
         else:
             # Science or SAA modes: handle observations and battery management
             self._handle_science_mode(utime, ra, dec, mode)
+
+    def _refresh_pointing_after_operations(
+        self,
+        previous_mode: ACSMode,
+        utime: float,
+        ra: float,
+        dec: float,
+        roll: float,
+        obsid: int,
+    ) -> tuple[float, float, float, int, ACSMode]:
+        mode = self.acs.get_mode(utime)
+        if mode == previous_mode and not self._has_due_acs_command(utime):
+            return ra, dec, roll, obsid, mode
+
+        ra, dec, roll, obsid = self.acs.pointing(utime)
+        self._sync_acs_slew_metadata()
+        return ra, dec, roll, obsid, self.acs.get_mode(utime)
+
+    def _has_due_acs_command(self, utime: float) -> bool:
+        command_queue = getattr(self.acs, "command_queue", None)
+        if not isinstance(command_queue, list):
+            return False
+        return any(command.execution_time <= utime for command in command_queue)
 
     def _handle_science_mode(
         self, utime: float, ra: float, dec: float, mode: ACSMode
--- a/conops/simulation/acs.py
+++ b/conops/simulation/acs.py
@@ -10,7 +10,8 @@ from ..common import (
     unixtime2date,
     unixtime2yearday,
 )
-from ..config import MissionConfig
+from ..common.vector import angular_separation
+from ..config import AttitudeConstraintPolicy, MissionConfig
 from ..simulation.passes import PassTimes
 from ..simulation.roll import optimum_roll
 from .acs_command import ACSCommand
@@ -572,6 +573,11 @@ class ACS:
         # Calculate roll angle (must run after _calculate_pointing so ra/dec are current).
         self.roll = self._compute_roll(utime)
 
+        # Idle is an executed attitude, not a constraint-free gap. If a completed
+        # observation is being held after science ends, move the hold to an
+        # attitude that satisfies the configured IDLE policy before recording.
+        self._enforce_idle_constraint_safe_attitude(utime)
+
         # Check current constraints (must run after roll is updated)
         self._check_constraints(utime)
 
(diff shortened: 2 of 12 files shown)


conops/simulation/acs.py
```diff
@@ -664,6 +670,202 @@ def _compute_roll(self, utime: float) -> float:
...
+            return
+
+        safe_attitude = self._find_constraint_safe_idle_attitude(utime, policy)
+        if safe_attitude is None:
+            msg = (
+                f"{unixtime2date(utime)}: No constraint-safe IDLE attitude found "
+                f"from RA={self.ra:.2f} Dec={self.dec:.2f}"
+            )
+            self._log_or_print(utime, "ERROR", msg)
+            raise RuntimeError(msg)
```
jak574: In general I'm against raising exceptions because of simulation failing to do something expected, as we're supposed to be simulating this very thing. This should be handled by fault management, not cause a crash. After all, this situation is presumably something that could happen on-orbit with a real spacecraft?

Here is a suggested change that logs the issue, and then force the spacecraft into SAFE. 
```suggestion
        if safe_attitude is None:
            cause = (
                f"No constraint-safe IDLE attitude found "
                f"(RA={self.ra:.2f} Dec={self.dec:.2f} policy={policy.value}); "
                f"requesting safe mode"
            )
            self._log_or_print(utime, "ERROR", f"{unixtime2date(utime)}: {cause}")
            fm = getattr(self.config, "fault_management", None)
            if fm is not None:
                fm.events.append(
                    FaultEvent(
                        utime=utime,
                        event_type="safe_mode_trigger",
                        name="idle_attitude_constraint",
                        cause=cause,
                        metadata={
                            "ra": self.ra,
                            "dec": self.dec,
                            "policy": policy.value,
                        },
                    )
                )
            self.request_safe_mode(utime)
            return
```

conops/ditl/queue_ditl.py
```diff
@@ -1378,6 +1381,29 @@ def _handle_mode_operations(
...
+
+        ra, dec, roll, obsid = self.acs.pointing(utime)
+        self._sync_acs_slew_metadata()
+        return ra, dec, roll, obsid, self.acs.get_mode(utime)
+
+    def _has_due_acs_command(self, utime: float) -> bool:
+        command_queue = getattr(self.acs, "command_queue", None)
+        if not isinstance(command_queue, list):
+            return False
+        return any(command.execution_time <= utime for command in command_queue)
```
jak574: Some unnecesary guarding here. `ACS.command_queue` is initialized as an empty list on `ACS` instantiation, so the `getattr` with `None` fallback is unnecessary, as the attribute always exists. 
```suggestion
    def _has_due_acs_command(self, utime: float) -> bool:
        return any(
            command.execution_time <= utime for command in self.acs.command_queue
        )
```

conops/simulation/acs.py
```diff
@@ -664,6 +670,202 @@ def _compute_roll(self, utime: float) -> float:
...
+        utime: float,
+        policy: AttitudeConstraintPolicy,
+    ) -> bool:
+        if policy == AttitudeConstraintPolicy.FULL_MISSION:
+            return self.constraint.in_constraint(
+                ra, dec, utime, target_roll=roll, acs_mode=ACSMode.IDLE
+            )
+        if policy == AttitudeConstraintPolicy.HARD_KEEPOUT:
+            return self._idle_attitude_violates_hard_keepout(ra, dec, roll, utime)
+        return False
```
jak574: This just returns false if `AttitudeConstraintPolicy` is not `FULL_MISSION` or `HARD_KEEPOUT`. Problem with this is that if we define another policy in the future, this behavior will be wrong, it should only return false if the value is `NONE`. So it's better to be explicit here:

```suggestion
        if policy == AttitudeConstraintPolicy.NONE:
            return False
        assert_never(policy)
```

`assert_never` which needs to be imported from `typing` will flag up if there's a value of `AttitudeConstraintPolicy` that isn't being captured by the if statements.

conops/simulation/acs.py
```diff
@@ -664,6 +670,202 @@ def _compute_roll(self, utime: float) -> float:
...
+            normalized = roll % 360.0
+            is_new = all(
+                abs(((normalized - existing + 180.0) % 360.0) - 180.0) > 1e-6
+                for existing in rolls
+            )
+            if is_new:
+                rolls.append(normalized)
+
+        add(optimal_roll)
+        for offset in range(0, 360, 15):
```
jak574: No need to start at 0 here, because that's covered by the previous line. Also `15` here is a magic number, define as a constant.

conops/simulation/acs.py
```diff
@@ -664,6 +670,202 @@ def _compute_roll(self, utime: float) -> float:
...
+            )
+            add(
+                (180.0 + float(self.ephem.sun_ra_deg[index])) % 360.0,
+                -float(self.ephem.sun_dec_deg[index]),
+            )
+        except (AttributeError, IndexError, TypeError, ValueError):
+            pass
+
+        grid: list[tuple[float, float]] = []
+        for dec in (-90, -75, -60, -45, -30, -15, 0, 15, 30, 45, 60, 75, 90):
```
jak574: Why use a tuple here instead of a range? Also same comment about 15 being hard coded as the step size.

conops/simulation/acs.py
```diff
@@ -664,6 +670,202 @@ def _compute_roll(self, utime: float) -> float:
...
+            if abs(dec) == 90:
+                grid.append((0.0, float(dec)))
+                continue
+            for ra in range(0, 360, 15):
```
jak574: 15 is a magic number.

jak574 (changes requested):

conops/simulation/acs.py
```diff
@@ -664,6 +675,223 @@ def _compute_roll(self, utime: float) -> float:
...
+        hold.slewdist = 0.0
+        hold.startra = ra
+        hold.startdec = dec
+        hold.startroll = roll
+        hold.endra = ra
+        hold.enddec = dec
+        hold.endroll = roll
+        hold.obstype = ObsType.IDLE
+        hold.obsid = IDLE_OBSID
+        hold.at = None
```
jak574: This is kind of bad form, and brittle code. I would put this into a `classmethod` in `Slew`, maybe as follows:
```python
    @classmethod
    def idle_hold(
        cls,
        config: MissionConfig,
        ra: float,
        dec: float,
        roll: float,
        utime: float,
    ) -> "Slew":
        """Create a zero-duration IDLE hold at the given attitude."""
        hold = cls(config=config)
        hold.slewrequest = utime
        hold.slewstart = utime
        hold.slewend = utime
        hold.startra = ra
        hold.startdec = dec
        hold.startroll = roll
        hold.endra = ra
        hold.enddec = dec
        hold.endroll = roll
        hold.obstype = ObsType.IDLE
        return hold
```

Then you can replace this code here with this:

```suggestion
        hold = Slew.idle_hold(self.config, ra, dec, roll, utime)
```

conops/simulation/acs.py
```diff
@@ -664,6 +675,223 @@ def _compute_roll(self, utime: float) -> float:
...
+            add(
+                (180.0 + float(self.ephem.earth_ra_deg[index])) % 360.0,
+                -float(self.ephem.earth_dec_deg[index]),
+            )
+            add(
+                (180.0 + float(self.ephem.sun_ra_deg[index])) % 360.0,
+                -float(self.ephem.sun_dec_deg[index]),
+            )
+        except (AttributeError, IndexError, TypeError, ValueError):
+            pass
```
jak574: I'm not really a fan of silent failures like these, and I'm not sure why you would need to guard like this.  Is this a mypy thing?

jak574 (changes requested):

jak574 (approved):
lgtm

PR #195, opened 2026-06-04
Restore cheap attitude-aware target slew scoring

Tohuvavohu:
## Summary

Restores cheap target-slew scoring without going back to the old endpoint-only estimate.

- Candidate ranking now estimates slew cost from quaternion attitude distance plus the configured ACS slew-time curve.
- Candidate ranking no longer builds a full `Slew` or materializes the 100-point slew path for every scored target.
- The scalar quaternion attitude distance is shared with the full `Slew` implementation, so the cheap scorer and full model cannot drift independently.
- The selected target still runs the full attitude-aware `Slew.calc_slewtime()` before enqueue, so execution, safety checks, and exported slew metadata continue to use the full slew model.
- Adds a regression test that fails if candidate scoring starts calling the full slew calculation again.
- Updates the target-scoring configuration docs.

## Tamalpais Accuracy / Runtime Assessment

Compared PR #190's full-estimator behavior against this branch with the Tamalpais `ConOps/plan_generation_config.yaml`, using the same stale cached EOP2 data and an in-memory run that did not write plan artifacts.

| Metric | Full estimator | Cheap quaternion estimator | Delta |
| --- | ---: | ---: | ---: |
| End-to-end elapsed | 145.65s | 28.81s | 5.1x faster |
| `QueueDITL.calc` | 145.36s | 28.40s | 5.1x faster |
| Science entries | 67 | 67 | 0 |
| Entry counts | AT 67 / CHARGE 15 / GSP 3 | AT 67 / CHARGE 15 / GSP 3 | identical |
| Total collection time | 62,820s | 62,820s | 0 |
| Total slew time | 12,950s | 12,950s | 0 |
| Ordered science obsid matches | 67/67 | 67/67 | identical |
| Unique science obsid Jaccard | 1.0 | 1.0 | identical |
| Aligned same-obsid slew delta | 0.0s mean / 0.0s max | 0.0s mean / 0.0s max | identical |

I also checked the simpler endpoint/great-circle cheap model before this patch. It was faster, but it materially changed target selection: only 30 common science obsids, Jaccard 0.316, and 1 ordered-position match. This branch keeps the schedule identical to the full estimator for the Tamalpais case while removing the repeated full path-generation cost.

## Validation

- `PYTHONPATH=/home/aaron/Projects/Cosmic/ops/coast-sim/.worktrees/cheap-target-slew-estimates /home/aaron/Projects/Cosmic/ops/coast-sim/coastvenv/bin/python -m pytest tests/vector/test_vector.py tests/slew/test_slew.py tests/target_queue/test_target_queue.py tests/scheduler_queue/test_queue_ditl.py -q`
  - `393 passed in 6.58s`
- `git diff --check`
  - passed
- `ruff check` and `ruff format --check` on affected files
  - passed
- Commit hooks:
  - available hooks passed: yaml check, EOF, whitespace, large-file check, ruff check, ruff format, pyupgrade
  - not run locally because executables are missing in this environment: `sphinx-build`, `mypy`


conops/common/vector.py
```diff
@@ -382,6 +382,22 @@ def attitude_to_quat(
...
+    roll1: float,
+    ra2: float,
+    dec2: float,
+    roll2: float,
+) -> float:
+    """Return the shortest angular distance between two attitudes in degrees."""
+    q1 = attitude_to_quat(ra1, dec1, roll1)
+    q2 = attitude_to_quat(ra2, dec2, roll2)
+    dot = abs(float(np.dot(q1, q2)))
+    dot = 1.0 if dot > 1.0 - 1e-15 else min(1.0, dot)
```
jak574: Everywhere else in the vector.py we handle the issue of values very close to zero or one in this way, rather than the asymetric approach above.
```suggestion
    dot = float(np.clip(abs(np.dot(q1, q2)), 0.0, 1.0))
```

jak574 (dismissed):
Goes some way towards fixing a large performance regression in the last few PRs. However things are still quite slow.

One minor fix and consider this approved if applied.

conops/common/vector.py
Tohuvavohu: Applied in facfa9f: switched quaternion_attitude_distance() to the np.clip(abs(np.dot(...)), 0.0, 1.0) pattern and kept the zero-distance regression test with an explicit micro-degree tolerance. Focused validation: 393 passed in 11.70s, ruff check/format passed, git diff --check passed.

PR #196, opened 2026-06-05
fix: cache charge deadline ephemeris timestamps

Tohuvavohu:
## Summary
- cache QueueDITL ephemeris timestamps as Unix floats for charge deadline lookups
- reuse the cached grid for snap-to-ephemeris and next-sunlight scans
- add regression coverage for cache reuse and refresh when ephem.timestamp is replaced

## Validation
- `python -m pytest tests/scheduler_queue/test_queue_ditl.py`
- `python -m ruff check conops/ditl/queue_ditl.py tests/scheduler_queue/test_queue_ditl.py`
- `python -m ruff format --check conops/ditl/queue_ditl.py tests/scheduler_queue/test_queue_ditl.py`
- pre-commit hooks passed during commit, including strict mypy
- Tamalpais plan-entry hash unchanged: `134fbebaf0c72b3c84913f24fd1bc501c8cb1c23f6e8c49bd9131ee37e729c66`

## Performance
On the current 24h / 500-target Tamalpais generation path, warmed timing improved from about `queue_ditl_calc=21.4s` to `queue_ditl_calc=13.1s`. This preserves the exact exported plan entries and only removes repeated timestamp conversion/scanning overhead in the charge-deadline path.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -124,6 +124,8 @@ class QueueDITL(DITLMixin, DITLStats):
         self.panel_power = list()
         # Eclipse tracking
         self.in_eclipse = list()
+        self._ephem_utime_cache: list[float] | None = None
+        self._ephem_utime_cache_source: Any | None = None
         # Subsystem power tracking
         self.power_bus = list()
         self.power_payload = list()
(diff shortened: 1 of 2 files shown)


jak574 (approved):
I'm not seeing the performance upgrade you're seeing, but there's nothing wrong with this per se.

PR #198, opened 2026-06-05
fix: use cheap pass slew trigger estimates

Tohuvavohu:
## Problem
While waiting for an upcoming ground-station pass, `QueueDITL` calls `Pass.time_to_slew()` on each scheduler tick to decide whether it is time to start slewing for the pass. That check only needs a scalar slew duration, but the old path built a full `Slew` and ran `Slew.calc_slewtime()`. In quaternion slew mode, that generates the 100-step quaternion slew path even though the path is thrown away immediately after the trigger decision.

On the Tamalpais 24h / 500-target plan this happened 991 times, costing about `2.6s` in the pass-trigger path. Actual commanded GSP slews still need full slew paths; this PR only removes full-path generation from the repeated trigger check.

## Summary
- compute the quaternion attitude distance and scalar slew time directly for ground-station pass trigger checks
- keep actual commanded GSP slews on the existing full `Slew.calc_slewtime()` path
- retain the full `Slew.calc_slewtime()` fallback for constraint-avoiding slew mode, where direct scalar equivalence is not guaranteed

## Validation
- `python -m pytest tests/passes/test_passes.py` (`59 passed`)
- `python -m pytest tests/scheduler_queue/test_queue_ditl.py` (`245 passed`)
- `python -m pytest tests/passes/test_passes.py tests/scheduler_queue/test_queue_ditl.py -q` (`304 passed`)
- `python -m ruff check conops/simulation/passes.py tests/passes/test_passes.py`
- `python -m ruff format --check conops/simulation/passes.py tests/passes/test_passes.py`
- `python -m mypy --strict --ignore-missing-imports conops/simulation/passes.py`
- commit hooks passed, including strict mypy

## Tamalpais Check
On the current 24h / 500-target Tamalpais generation path after #196, this branch produced the same plan-entry hash:

`134fbebaf0c72b3c84913f24fd1bc501c8cb1c23f6e8c49bd9131ee37e729c66`

Observed warmed timing improved from roughly `queue_ditl_calc=13.4s` post-#196 to `queue_ditl_calc=10.6-11.0s` on this branch. The pass trigger path dropped from about `2.6s` to about `0.02s` in the targeted profile, and full `Slew.calc_slewtime()` calls dropped from 1251 to 260.

### DIFF
--- a/conops/simulation/passes.py
+++ b/conops/simulation/passes.py
@@ -331,6 +331,9 @@ class Pass(BaseModel):
             raise ValueError("ACS config must be set to calculate slew time")
 
         acs_config = self.config.spacecraft_bus.attitude_control
+        # This scalar shortcut is equivalent to Slew.calc_slewtime() only for
+        # quaternion slews. Constraint-avoiding and future slew algorithms keep
+        # the full path unless their scalar equivalence has been proven.
         if acs_config.slew_algorithm == SlewAlgorithm.QUATERNION:
             slewdist = quaternion_attitude_distance(
                 ra,
--- a/conops/simulation/passes.py
+++ b/conops/simulation/passes.py
@@ -7,10 +7,11 @@ import rust_ephem
 from pydantic import BaseModel, Field
 
 from ..common import ics_date_conv, unixtime2date
-from ..common.enums import AntennaType, ObsType
+from ..common.enums import AntennaType, ObsType, SlewAlgorithm
 from ..common.vector import (
     attitude_for_body_vector_tracking,
     body_vector_to_eci,
+    quaternion_attitude_distance,
     radec2vec,
     rotvec,
     separation,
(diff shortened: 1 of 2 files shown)


conops/simulation/passes.py
```diff
@@ -329,6 +330,26 @@ def _slew_time_to_target(
...
+                roll,
+                target_ra,
+                target_dec,
+                target_roll,
+            )
+            if np.isnan(slewdist) or slewdist < 0:
+                raise ValueError(
+                    f"Invalid slew distance: {slewdist} "
+                    f"(start={ra},{dec} end={target_ra},{target_dec})"
+                )
```
jak574: Don't think this is needed. Even if it is, it's not great to have just a crash here without resolution, but I genuinely can't see how this is triggered.

jak574 (changes requested):

conops/simulation/passes.py
Tohuvavohu: Removed this guard in acf7e99. Agreed it should not be reachable for quaternion_attitude_distance(), and adding a new hard failure in the trigger estimate path was not useful.

jak574 (approved):
lgtm

PR #199, opened 2026-06-05
fix: skip impossible target candidates before slew estimates

Tohuvavohu:
## Summary
- add a zero-slew impossibility prefilter in TargetQueue.get() before candidate slew estimation
- skip candidates only when the current immediate visibility/deadline rules cannot fit ss_min even with zero slew
- keep collection-deadline prefiltering limited to the scored path so the existing unscored fast path behavior is preserved

## Safety
This is a rejection-only filter. Under the current TargetQueue semantics, science collection for a candidate cannot start before utime and real slew times are nonnegative. If the target cannot fit the required ss_min window with zero slew, adding any real slew cannot make that same immediate visibility/deadline window feasible.

The helper is intentionally named and documented as a zero-slew impossibility check, not a ranking heuristic. If future scheduling semantics allow waiting for future visibility before target selection or collecting before target acquisition, this helper should be revisited.

## Validation
- ruff format conops/targets/target_queue.py tests/target_queue/test_target_queue.py
- ruff check conops/targets/target_queue.py tests/target_queue/test_target_queue.py
- pytest tests/target_queue/test_target_queue.py
- pytest tests/scheduler_queue/test_queue_ditl.py::TestFetchNewPPT
- Tamalpais repeatable generation check: repeatable=true, plan_entries=85, attitude_samples=1440
- Normalized Tamalpais plan hash matches origin/main: d78f4544f7a8935f9e481b0efbf80bf13ccef1f48c2ad5648fec0221a6966cf3

## Timing Sample
Single-run Tamalpais queue_ditl_calc samples on this machine:
- origin/main f9c5574: 15.317s, 15.561s
- this branch: 14.131s, 14.242s

These are coarse wall-clock samples, but the normalized plan payload is unchanged.

### DIFF
--- a/conops/targets/target_queue.py
+++ b/conops/targets/target_queue.py
@@ -150,9 +150,12 @@ class TargetQueue:
         visibility_window: list[float],
         utime: float,
         deadline: float | None = None,
+        slewtime: float | None = None,
     ) -> float:
         """Estimate useful science collection available after the slew."""
-        collection_start = utime + float(target.slewtime)
+        collection_start = utime + float(
+            target.slewtime if slewtime is None else slewtime
+        )
         collection_end = float(visibility_window[1])
         if deadline is not None:
             collection_end = min(collection_end, float(deadline))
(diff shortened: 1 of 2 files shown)


jak574 (approved):
lgtm

PR #201, opened 2026-06-05
fix: prune target scoring with score upper bound

Tohuvavohu:
## Summary
- add a conservative candidate score upper bound before expensive slew estimation
- skip scored candidates when their optimistic zero-slew/no-penalty score cannot beat the current best score
- keep the old full scoring path for negative weights, because negative penalties can reward costs

## Safety
This is exact branch-and-bound pruning, not approximate top-N selection. The bound assumes zero slew and ignores all nonnegative penalties, so it is at least as high as the candidate real score whenever the configured weights are nonnegative. Candidates are skipped only when that optimistic score is still <= the current best score; equal scores are safe to skip because TargetQueue already keeps the first best score and only replaces on strictly greater score.

If any scoring weight is negative, pruning is disabled and the previous full scoring behavior is preserved.

## Validation
- ruff format conops/targets/target_queue.py tests/target_queue/test_target_queue.py
- ruff check conops/targets/target_queue.py tests/target_queue/test_target_queue.py
- pytest tests/target_queue/test_target_queue.py
- Tamalpais repeatable generation: repeatable=true, plan_entries=85, attitude_samples=1440
- Revalidated after #197 landed: normalized Tamalpais plan hash matches current origin/main a7bad5a and PR #201-on-current-main: 6d7018bcec7f23b77e1ad205bde0a915da396d0e526d1906b98649e63f2b6d75

## Timing Sample
Single-run Tamalpais queue_ditl_calc samples on this machine:
- current origin/main a7bad5a: 9.663s
- PR #201 applied on current main: 8.353s

These are coarse wall-clock samples, but the normalized plan payload is unchanged.

### DIFF
--- a/conops/targets/target_queue.py
+++ b/conops/targets/target_queue.py
@@ -199,6 +199,47 @@ class TargetQueue:
         )
         return collection_seconds >= float(target.ss_min)
 
+    def _score_upper_bound_pruning_is_safe(self) -> bool:
+        """Return True when configured weights make the score bound conservative."""
+        return (
+            self.slew_distance_weight >= 0.0
+            and self.slew_time_weight >= 0.0
+            and self.collection_time_weight >= 0.0
+            and self.radiator_sun_exposure_weight >= 0.0
+            and self.radiator_earth_exposure_weight >= 0.0
+        )
+
+    def _candidate_score_upper_bound(
+        self,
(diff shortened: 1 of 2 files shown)


jak574 (approved):
lgtm

PR #202, opened 2026-06-05
Add plan metadata schema support

Tohuvavohu:
## Summary
- add a first-class `metadata` dict to `Plan`
- serialize and load top-level plan metadata through `PlanSchema`
- omit empty metadata during file save to avoid changing generic plan exports

## Why
Mission-specific plan generators need a supported place for provenance like selected TLE source data and derived validation values. Without this field, downstream generators have to reopen the saved JSON and append unmodeled keys that `Plan.load()` does not preserve.

## Scope
This PR intentionally does not include TLE-to-classical-element conversion. That reusable orbital/TLE capability is now covered by rust-ephem PR #173. Coast only provides the plan metadata container; mission/config code can populate it with rust-ephem-derived values after #173 lands.

## Validation
- `coastvenv/bin/python -m pytest tests/plan/test_plan.py tests/plan/test_plan_schema.py`
- `coastvenv/bin/python -m ruff check conops/common/__init__.py conops/targets/plan.py conops/targets/plan_schema.py tests/plan/test_plan.py tests/plan/test_plan_schema.py`
- `coastvenv/bin/python -m ruff format --check conops/common/__init__.py conops/targets/plan.py conops/targets/plan_schema.py tests/plan/test_plan.py tests/plan/test_plan_schema.py`
- `git diff --check`


jak574 (changes requested):
TLE stuff should live in rust_ephem IMHO. I put together a PR here:

https://github.com/CosmicFrontierLabs/rust-ephem/pull/173

jak574 (approved):
I pushed a change that bumps the minimum version of rust-ephem to the version that supports TLE properties. LGTM

PR #204, opened 2026-06-09
Add UHF communications band support

Tohuvavohu:
## Summary

- allow UHF as a supported BandCapability value
- keep UHF rates explicit instead of adding generic defaults
- document UHF support and add tests for parsing/defaults and effective downlink matching

## Why

Some mission ground-station RF configurations use UHF as the ground receive band. Without UHF in the COAST schema, mission configs either fail validation or silently omit usable low-rate contacts.

## Validation

- coastvenv/bin/python -m pytest tests/config/test_communications.py tests/ditl/test_ditl_mixin_coverage.py tests/ditl/test_ditl_data_rate.py
- coastvenv/bin/python -m ruff check conops/config/communications.py tests/config/test_communications.py tests/ditl/test_ditl_mixin_coverage.py
- coastvenv/bin/python -m ruff format --check conops/config/communications.py tests/config/test_communications.py tests/ditl/test_ditl_mixin_coverage.py

### DIFF
--- a/conops/config/communications.py
+++ b/conops/config/communications.py
@@ -12,7 +12,7 @@ from ._base import ConfigModel
 class BandCapability(ConfigModel):
     """Defines a communication band's data rate capabilities."""
 
-    band: Literal["S", "X", "Ka", "Ku", "L", "C", "K"]  # Common spacecraft bands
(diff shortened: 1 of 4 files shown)


jak574 (approved):
lgtm

PR #203, opened 2026-06-09
Add TLE plan metadata helper

Tohuvavohu:
## Summary

- add tle_plan_metadata() to build generic TLE-backed plan metadata
- add attach_tle_plan_metadata() to attach that metadata to a Plan before schema export
- delegate classical element derivation to rust-ephem TLERecord.classical_elements()

## Why

Mission repos should not assemble COAST plan metadata JSON by hand or duplicate orbital derivation logic. rust-ephem owns the TLE/COE derivation and coast-sim owns the plan metadata shape/serialization; mission config code only selects the TLE and calls this helper.

## Validation

- coastvenv/bin/python -m pytest tests/plan/test_plan_metadata.py tests/plan/test_plan.py tests/plan/test_plan_schema.py
- coastvenv/bin/python -m ruff check conops/targets/plan_metadata.py conops/targets/__init__.py tests/plan/test_plan_metadata.py
- coastvenv/bin/python -m ruff format --check conops/targets/plan_metadata.py conops/targets/__init__.py tests/plan/test_plan_metadata.py

PR #207, opened 2026-06-09
Fix/tle plan metadata helper pydantic

jak574:
I don't like the implementation in #203. Here's a version that uses Pydantic. Please let me know if this meets requirements.

### DIFF
--- a/conops/targets/__init__.py
+++ b/conops/targets/__init__.py
@@ -2,7 +2,6 @@ from .plan import Plan, TargetList
 from .plan_entry import PlanEntry
 from .plan_metadata import (
     PlanMetadata,
-    PlanSchemaMetadata,
     TLEEphemerisMetadata,
 )
 from .plan_schema import (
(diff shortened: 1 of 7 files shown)


Tohuvavohu:
@jak574  The optional fields fix addresses the partial-TLE example, thanks. I think there is still one generic-metadata issue to decide before merge.

If `metadata.ephemeris` is intended to be a generic ephemeris/provenance slot, non-TLE payloads should preserve their fields. Right now a payload like:

```python
{"ephemeris": {"source": "SPICE", "kernel": "example.bsp"}}
```

validates, but serializes back as a TLE-shaped object: `kernel` is dropped and the TLE classical-elements note is added. That is surprising for non-TLE metadata.

My preference: keep top-level `metadata` generic and preserve unknown/non-TLE `metadata.ephemeris` payloads. If we want a strict typed TLE model, put it under a narrower key like `metadata.tle_ephemeris` or `metadata.ephemeris.tle`, or use a union/discriminator so only TLE source payloads get coerced into `TLEEphemerisMetadata`. Also, the TLE note should only be emitted for TLE metadata generated by `PlanMetadata.from_tle_record(...)`, not for arbitrary ephemeris metadata.

For the immediate TLE/COE requirement this PR is close; this is mostly about avoiding COAST taking over `metadata.ephemeris` globally in a way that loses other mission metadata.

PR #209, opened 2026-06-10
fix: Enforce full-mission attitude constraints

Tohuvavohu:
## Summary
- Default ACS modes to the full-mission attitude constraint policy unless a mode opts into a narrower policy.
- Filter ground-station pass profiles against PASS attitude constraints before scheduling.
- Preflight science, charge, and ground-station slew paths against the relevant ACS-mode constraints at ephemeris-grid samples.
- Align combined star-tracker planning constraints with the per-tracker telemetry roll convention so planning and executed telemetry evaluate the same STR body vectors.
- Log constraint-dropped ground-station pass opportunities even when no passes survive filtering.
- Add tests covering mode policy defaults, pass filtering, slew rejection paths, and nonzero-roll STR planning/telemetry consistency.

## Validation
- `PYTHONPATH=. coastvenv/bin/python -m pytest tests/star_tracker tests/constraint` -> 167 passed.
- `PYTHONPATH=. coastvenv/bin/python -m pytest tests/scheduler_queue tests/passes` -> 390 passed.
- `PATH=coastvenv/bin:$PATH pre-commit run --files conops/config/star_tracker.py tests/star_tracker/test_star_tracker.py` -> passed.
- `PATH=coastvenv/bin:$PATH pre-commit run --files conops/ditl/queue_ditl.py tests/scheduler_queue/test_queue_ditl.py` -> passed.
- End-to-end tam-config plan generation completed in about 11 seconds on the current branches.
- Regenerated visualizer telemetry reports 0 `in_constraint` samples, 0 STR hard samples, 0 STR soft samples, and 0 samples with fewer than 2 functional star trackers.

## Notes
This is paired with tamalpais-configuration PR #21, which sets all configured ACS modes to full-mission constraints and regenerates the plan. The safety check is sampled on COAST's existing ephemeris grid; it is not a continuous between-sample proof.

PR #210, opened 2026-06-11
feat: Plan constraint-safe ground station roll phases

Tohuvavohu:
## Problem

Ground-station passes use a fixed body-frame antenna, so the spacecraft has one remaining attitude degree of freedom while tracking a station: roll about the station line of sight.

Before this PR, GSP attitude generation picked that roll from the default reference-vector convention. That preserved antenna pointing, but it did not search for a constraint-safe roll. With full-mission PASS attitude constraints enabled, a geometrically visible station pass could be dropped even when another roll about the same antenna line of sight would satisfy the attitude policy.

The current Tamalpais case showed this directly: Svalbard had two viable tracking opportunities, but one was rejected because no single default/fixed roll phase stayed constraint-safe for the whole contact.

## Solution

This PR makes GSP tracking roll an explicit planning variable.

- Generate candidate roll phases around the fixed antenna-to-station line of sight.
- Prefer the legacy/default phase first, then nearby phases, so existing behavior is preserved when it is already safe.
- Control the roll phase search granularity with `spacecraft_bus.attitude_control.gsp_tracking_phase_step_deg` (default `5.0` deg), so users can trade CPU cost against search resolution.
- First try to find one fixed phase that is safe for the whole pass.
- If no fixed phase works, fall back to a time-varying phase path across the pass samples.
- Accept a dynamic path only when every sampled attitude satisfies the PASS attitude policy and every adjacent sample-to-sample attitude change fits within ACS `motion_time()`.
- If no safe feasible path exists, keep dropping the pass and retain the default profile for diagnostics.

This keeps antenna pointing intact: phase changes only rotate the spacecraft about the station line of sight.

## Safety Bound

The guarantee is aligned with COAST discrete ephemeris-grid validation model:

- sampled GSP attitudes are checked against the configured PASS attitude policy;
- dynamic roll transitions are checked for ACS motion feasibility between adjacent samples;
- pass slews are still separately preflighted by the existing slew path constraint checks.

This does not claim a continuous-time proof between ephemeris samples. It makes GSP roll planning as safe as the current COAST pass/slew validation framework can verify.

## Validation

- `PYTHONPATH=. MPLCONFIGDIR=/tmp coastvenv/bin/python -m pytest tests/config/test_config.py -k 'gsp_tracking_phase_step'`
- `PYTHONPATH=. MPLCONFIGDIR=/tmp coastvenv/bin/python -m pytest tests/passes/test_passes.py`
- `coastvenv/bin/ruff check conops/config/acs.py conops/simulation/passes.py tests/config/test_config.py tests/passes/test_passes.py`
- `coastvenv/bin/ruff format --check conops/config/acs.py conops/simulation/passes.py tests/config/test_config.py tests/passes/test_passes.py`
- Previous non-exporting Tamalpais DITL: 2 SVAL GSP entries, 0 dropped passrequests, 0 constraint telemetry hits, 0 plan/execution mismatches

## Notes

- Depends on #209 / `feature/full-mission-attitude-constraints`.
- If #209 merges first, retarget this PR to `main`.

### DIFF
--- a/conops/simulation/passes.py
+++ b/conops/simulation/passes.py
@@ -613,12 +613,12 @@ class PassTimes:
             attitude_profile.append(attitude)
         return attitude_profile
 
-    def _constraint_safe_tracking_attitude_profile(
+    def _fixed_phase_tracking_attitude_profile(
         self,
         antenna_boresight: tuple[float, float, float],
         target_vectors: np.ndarray,
         track_utime: list[float],
-    ) -> tuple[list[tuple[float, float, float]], bool] | None:
+    ) -> tuple[list[tuple[float, float, float]] | None, bool]:
         fallback_profile: list[tuple[float, float, float]] | None = None
         for phase_deg in self._gsp_tracking_phase_candidates():
             attitude_profile = self._tracking_attitude_profile_for_phase(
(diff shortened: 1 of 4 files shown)


conops/simulation/passes.py
```diff
@@ -29,6 +29,7 @@
 # Legacy ground-pass roll for profiles without explicit roll samples.
 GSP_TRACK_ROLL = 0.0
 LEGACY_GSP_ANTENNA_BODY_VECTOR = (-1.0, 0.0, 0.0)
+GSP_TRACK_PHASE_STEP_DEG = 5.0
```
jak574: Can this be made into a configuration parameter rather than a constant? Maybe move it into `AttitudeControlSystem` config? This way the user has control over how much CPU they want to expend vs the granularity of the search.

jak574 (changes requested):

jak574 (approved):
lgtm

PR #213, opened 2026-06-20
fix: log recharge-threshold interruptions as charging

Tohuvavohu:
## Summary
- classify recharge-threshold science interruptions as `CHARGING` events instead of `ERROR`
- replace the misleading `BATTERY ALERT` / `emergency charging` wording with `Battery below recharge threshold`
- keep scheduling behavior unchanged: recharge still starts when `battery_alert` is true and sunlight is available
- add tests for both the active `QueueDITL` path and the legacy `EmergencyCharging` wrapper

## Why
Plans were reporting `ERROR` events even though battery state of charge stayed well above the depth-of-discharge safety floor. The root cause is that `battery_alert` fires below the recharge threshold, currently 95%, not only near the safety floor. Those interruptions are normal recharge-window operations, so they should not inflate plan error counts.

### DIFF
--- a/conops/simulation/emergency_charging.py
+++ b/conops/simulation/emergency_charging.py
@@ -86,14 +86,26 @@ class EmergencyCharging:
         self._charging_suppressed_due_to_eclipse = False
         self.log = log
 
-    def _log_or_print(self, utime: float, event_type: str, description: str) -> None:
+    def _log_or_print(
+        self,
+        utime: float,
+        event_type: str,
+        description: str,
+        obsid: int | None = None,
+        acs_mode: ACSMode | None = None,
+    ) -> None:
         """Log event to DITLLog if available, otherwise print."""
         if self.log is not None:
             self.log.log_event(
-                utime=utime, event_type=event_type, description=description
+                utime=utime,
+                event_type=event_type,
+                description=description,
+                obsid=obsid,
+                acs_mode=acs_mode,
             )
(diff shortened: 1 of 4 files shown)


conops/simulation/emergency_charging.py
```diff
@@ -187,8 +187,9 @@ def initiate_emergency_charging(
         if current_ppt is not None and not getattr(current_ppt, "done", False):
             self._log_or_print(
                 utime,
-                "ERROR",
-                "BATTERY ALERT: Terminating science observation for emergency charging",
+                "CHARGING",
+                "Battery below recharge threshold; interrupting science observation "
+                "for charging",
```
jak574, line 204: So big issue here is that we have two places where this same error is reported, here and in queue_ditl.py, but this one doesn't report `obsid` value and the one in `queue_ditl.py` does, meaning that the error will be inconsistent.

jak574 (changes requested):

conops/simulation/emergency_charging.py
Tohuvavohu: Addressed in 29957db by passing the interrupted science obsid through EmergencyCharging._log_or_print into DITLLog.log_event. The existing battery test now asserts the CHARGING event carries the interrupted obsid, matching the queue_ditl path.

jak574 (approved):

PR #214, opened 2026-06-21
Use explicit attitude constraint scopes

Tohuvavohu:
## Problem
The previous policy model was too coarse. It could distinguish science from safety, but it could not express the actual operational categories independently:

- hardware safety keepouts that apply to all executed attitudes
- imaging-quality constraints used for science collection and target selection
- power-generation constraints used for charge and power-positive attitudes
- ground-contact constraints used for GSP tracking attitudes

That made charge and pass modes either inherit imaging-quality rules they should not enforce, or skip non-imaging constraints they still need.

## Solution
- Replace `AttitudeConstraintPolicy` / `attitude_constraint_policy` with explicit `AttitudeConstraintScope` / `attitude_constraint_scopes`.
- Add scopes: `hardware_safety`, `imaging_quality`, `power_generation`, and `ground_contact`.
- Default mode scopes:
  - `SCIENCE`: hardware safety + imaging quality + power generation
  - `CHARGING`: hardware safety + power generation
  - `SLEWING`: hardware safety
  - `PASS`: hardware safety + power generation + ground contact
  - `SAA`: hardware safety
  - `SAFE`: hardware safety + power generation
  - `IDLE`: hardware safety
- Classify top-level Sun/Earth/Moon/orbit/anti-sun/STR-soft/science constraints as imaging quality.
- Classify panel constraints as power generation.
- Classify explicit safety, STR-hard, radiator-hard, and telescope-hard constraints as hardware safety.
- Add a separate `ground_contact_constraint` hook for contact tracking constraints.
- Route ACS idle holds, emergency charging, GSP pass filtering, housekeeping telemetry, locked-roll science checks, and post-run validation through configured scopes.
- Remove the old policy compatibility aliases and old `science_plus_safety` naming from the touched API surface.

## Compatibility
This is intentionally a breaking config API change. Mission configs should use `attitude_constraint_scopes` with lists of scope strings. Old `attitude_constraint_policy` values are not preserved.

## Validation
- `PYTHONPATH=. coastvenv/bin/python -m pytest tests/config/test_config.py tests/constraint/test_constraint.py tests/acs/test_acs.py tests/acs/test_acs_roll.py tests/passes/test_passes.py tests/scheduler_queue/test_queue_ditl.py -q`
  - `496 passed`
- `git diff --check`
  - clean
- Verified with the matching Tamalpais config update in CosmicFrontierLabs/tamalpais-configuration#23.

### DIFF
--- a/conops/config/constraint.py
+++ b/conops/config/constraint.py
@@ -28,12 +28,14 @@ _TIME_PRECISION = 0  # round to nearest second
 _RA_DEC_ROUNDER = 10**_RA_DEC_PRECISION
 _TIME_ROUNDER = 10**_TIME_PRECISION
 
-_POLICY_SCIENCE_PLUS_SAFETY = "science_plus_safety"
-_POLICY_SCIENCE = "science"
-_POLICY_SAFETY = "safety"
-_POLICY_FULL_MISSION = "full_mission"
-_POLICY_HARD_KEEPOUT = "hard_keepout"
-_POLICY_NONE = "none"
+
+class AttitudeConstraintScope(str, Enum):
+    """Constraint scope applied to executed attitude samples."""
+
+    HARDWARE_SAFETY = "hardware_safety"
+    IMAGING_QUALITY = "imaging_quality"
+    POWER_GENERATION = "power_generation"
+    GROUND_CONTACT = "ground_contact"
 
 
 def _default_sun_constraint() -> ConstraintConfig:
@@ -124,14 +126,14 @@ class Constraint(ConfigModel):
     )
     panel_constraint: ConstraintConfig | None = Field(
         default=None,
-        description="Solar panel constraint configuration",
+        description="Power-generation / solar panel constraint configuration",
     )
     science_constraint: ConstraintConfig | None = Field(
         default=None,
         description=(
-            "Additional science-quality / scheduling constraint. Legacy top-level "
-            "sun, earth, moon, orbit, panel, anti-sun, and star-tracker soft "
-            "constraints are also treated as science constraints."
+            "Additional imaging-quality / science scheduling constraint. Top-level "
+            "sun, earth, moon, orbit, anti-sun, and star-tracker soft constraints "
+            "are also treated as imaging-quality constraints."
         ),
     )
     safety_constraint: ConstraintConfig | None = Field(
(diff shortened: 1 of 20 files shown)


conops/config/constraint.py
```diff
@@ -116,7 +126,27 @@ class Constraint(ConfigModel):
...
+        default=None,
+        description=(
+            "Additional hardware-safety constraint. Star-tracker hard, radiator "
+            "hard, and telescope hard constraints are also treated as safety "
+            "constraints."
+        ),
+    )
+    ground_contact_constraint: ConstraintConfig | None = Field(
+        default=None,
+        description="Additional ground-contact attitude constraint",
```
jak574: Description is a little terse. How's this?
```suggestion
        description=(
            "Keepout constraint applied during ground-contact (PASS) attitudes. "
            "Violated (returns True) when the pointing direction would interfere "
            "with ground-station contact, e.g. an Earth-limb or antenna-exclusion "
            "zone. Included in the 'ground_contact' scope."
        ),
```

conops/config/constraint.py
```diff
@@ -300,13 +389,28 @@ def roll_independent_constraint(self) -> ConstraintConfig | None:
...
-        active = [c for c in components if c is not None]
-        if not active:
-            return None
-        combined = active[0]
-        for c in active[1:]:
-            combined = combined | c
-        return combined
+        components.extend(
+            constraint
+            for constraint in (self.science_constraint, self.safety_constraint)
```
jak574: Should add `self.ground_contact_constraint` here?

```suggestion
            for constraint in (
                self.science_constraint,
                self.safety_constraint,
                self.ground_contact_constraint,
            )
```

conops/config/constraint.py
jak574, line 1530: I think we should add `in_ground_constraint` to the list to the list of constraints tracked here:

```python
        if self.in_ground_contact(ra=ra, dec=dec, time=time, target_roll=target_roll):
            count += 2
```

jak574 (changes requested):
A few minor fixes.

conops/config/constraint.py
Tohuvavohu: Applied this wording in a644386.

conops/config/constraint.py
Tohuvavohu: Added ground_contact_constraint to the roll-independent aggregate in a644386, with a focused test covering a roll-independent ground-contact constraint.

conops/config/constraint.py
Tohuvavohu: Added in_ground_contact() to in_constraint_count() in a644386, with a focused count test for a ground-contact-only violation.

jak574 (approved):
lgtm

PR #215, opened 2026-06-24
fix: scope ACS constraint logging

Tohuvavohu:
## Summary
- Make ACS legacy `CONSTRAINT` telemetry use `attitude_constraint_name_for_scopes()` with the current ACS mode scopes.
- Preserve the `CONSTRAINT` event type for compatibility, but include the active scope label in the message.
- Add regressions for `IDLE`/`hardware_safety` ignoring panel violations and `CHARGING`/`power_generation` reporting them.

## Problem
After explicit attitude constraint scopes were added, planning and validation used scoped attitude constraints, but `ACS._check_constraints()` still sampled the global `in_constraint()` helper. That made `IDLE` attitudes log `Panel` violations even when `IDLE` only enforced `hardware_safety`, which confused downstream mission seed gates.

## Validation
- `PYTHONPATH=. coastvenv/bin/python -m pytest tests/acs/test_acs.py tests/constraint/test_constraint.py -q`
- `coastvenv/bin/python -m ruff check conops/simulation/acs.py tests/acs/test_acs.py`
- `git diff --check`

Note: local pre-commit stopped only because the `mypy` executable is unavailable in this checkout; ruff and tests passed.

### DIFF
--- a/conops/simulation/acs.py
+++ b/conops/simulation/acs.py
@@ -13,6 +13,7 @@ from ..common import (
 from ..common.vector import angular_separation
 from ..config import AttitudeConstraintScope, FaultEvent, MissionConfig
 from ..config.constraint import (
+    attitude_constraint_name_for_scopes,
     attitude_constraint_scope_label,
     in_attitude_constraint_scopes,
 )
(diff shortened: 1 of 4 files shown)


conops/simulation/acs.py
```diff
@@ -911,45 +912,32 @@ def _check_constraints(self, utime: float) -> None:
...
+            scopes = self.config.attitude_constraint_scopes_for_mode(self.acsmode)
+            constraint_name = attitude_constraint_name_for_scopes(
+                self.constraint,
+                scopes,
                 self.last_slew.at.ra,
                 self.last_slew.at.dec,
                 utime,
                 target_roll=self.roll,
+                acs_mode=self.acsmode,
             )
```
jak574, line 927: This is a regression, as before it would report *every* violated constraint, now it only reports only the first one.

jak574 (changes requested):
One issue

jak574 (approved):
lgtm

PR #216, opened 2026-06-25
fix: export scoped attitude constraint telemetry

Tohuvavohu:
## Problem
The DITL housekeeping `in_constraint` field reports any globally configured constraint hit. That is useful diagnostic data, but it is not scoped by the ACS mode's active attitude constraint scopes. Downstream consumers can therefore interpret a non-active science/performance keepout as an executed hard violation.

## Solution
- Add `Housekeeping.attitude_constraint` for the active scoped attitude constraint violation, if any.
- Add `Housekeeping.attitude_constraint_scope` with the configured scope label for that active violation.
- Keep legacy `in_constraint` unchanged for compatibility and diagnostics.
- Populate the new fields from the same scoped result used by post-simulation validation.

### DIFF
--- a/conops/ditl/ditl.py
+++ b/conops/ditl/ditl.py
@@ -8,6 +8,11 @@ from conops.targets.plan import Plan
 from ..common import angular_separation, dtutcfromtimestamp, radec2vec, scbodyvector
 from ..common.vector import attitude_to_quat
 from ..config import MissionConfig
+from ..config.constraint import (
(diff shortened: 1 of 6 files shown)


jak574 (changes requested):
The extra parameters are only added to `HouseKeeping` in`QueueDITL` it should also be added to `DITL.calc()` for consistency.

jak574 (approved):
lgtm

PR #217, opened 2026-06-25
Add battery recharge hysteresis

Tohuvavohu:
## Problem

Emergency charging used a single `recharge_threshold` both to start and clear the alert. In the TAM schedule this lets COAST stop charging as soon as state of charge reaches 95%, schedule one short payload interval, then immediately fall below 95% again and re-enter charging. In the visualizer this shows up as gaps/slew churn in the middle of what should be a continuous Sun-charge recovery.

## Solution

Add a latched recharge clear threshold:

- `recharge_threshold` starts emergency recharge.
- `recharge_clear_threshold` clears emergency recharge.
- If unspecified, `recharge_clear_threshold` defaults to `recharge_threshold + 0.005`, capped at 1.0.

While `emergency_recharge` is active, `battery_alert` now stays true until the higher clear threshold is reached. This keeps the scheduler in charging mode through the small SOC buffer instead of bouncing around the lower threshold.

## Why 0.5%

I tested a larger 2% default first. It removed the chatter, but it was too conservative for the current TAM plan: charge time increased to 184 minutes and science fell to 1210 minutes. A 0.5% default still removes the immediate restart pattern while keeping the cost smaller:

- 92 entries total
- 75 AT entries, 15 CHARGE entries, 2 GSP entries
- 1262 minutes AT
- 137 minutes CHARGE
- 24 minutes GSP
- 0 CHARGE-to-CHARGE gaps
- 0 charge restarts within 3 minutes

### DIFF
--- a/conops/config/battery.py
+++ b/conops/config/battery.py
@@ -28,7 +28,15 @@ class Battery(ConfigModel):
         default=0.3, description="Maximum allowed depth of discharge (0.0-1.0)"
     )
     recharge_threshold: float = Field(
-        default=0.95, description="Battery level fraction to stop charging (0.0-1.0)"
+        default=0.95,
+        description="Battery level fraction that starts emergency recharge (0.0-1.0)",
+    )
+    recharge_clear_threshold: float | None = Field(
+        default=None,
+        description=(
+            "Battery level fraction that clears emergency recharge. Defaults to "
+            "recharge_threshold + 0.005, capped at 1.0."
+        ),
     )
     charge_level: float = Field(
         default=0, description="Current battery charge level (Wh)"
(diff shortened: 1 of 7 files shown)


conops/config/battery.py
```diff
@@ -72,18 +96,27 @@ def charge_state(self) -> ChargeState:
     @property
     def battery_alert(self) -> bool:
         """Is the battery in an alert status caused by discharge"""
+        threshold_tolerance = 1e-12
```
jak574: 1e-12 is a magic number

conops/config/battery.py
```diff
@@ -45,8 +53,24 @@ def set_defaults(cls, values: dict[str, Any]) -> dict[str, Any]:
         if "watthour" not in values:
             values["watthour"] = values.get("amphour", 20) * values.get("voltage", 28)
 
+        if values.get("recharge_clear_threshold") is None:
+            recharge_threshold = float(values.get("recharge_threshold", 0.95))
+            values["recharge_clear_threshold"] = min(1.0, recharge_threshold + 0.005)
```
jak574: 0.95 and 0.005 are magic numbers

jak574 (changes requested):
Fix magic number usage.

jak574 (approved):
lgtm

PR #218, opened 2026-06-26
Export GSP station coordinates

Tohuvavohu:
## Summary
- add station_lat_deg, station_lon_deg, and station_alt_m to exported GSP plan entries
- populate the fields from the configured ground-station registry when a GSP plan entry is recorded
- cover PlanSchema round-trip and QueueDITL GSP export behavior


jak574 (approved):
lgtm

PR #227, opened 2026-07-26
Fix annotated YAML round trips for unnamed list models

Tohuvavohu:
## Problem

The annotated YAML writer assumes every model inside a list has a `name` field. For models such as `BandCapability`, it writes a synthetic `name: null` key. With strict unknown-field validation, Coast can therefore generate YAML that its own reader rejects.

## Changes

- Serialize list-model items from their declared schema fields instead of inventing a `name` key.
- Preserve `name` as the leading field for models that actually declare it.
- Keep strict `extra="forbid"` behavior unchanged; existing invalid files must be regenerated or explicitly migrated.
- Add communications round-trip regression coverage.

## Validation

- `pytest tests/config/test_config.py tests/config/test_communications.py`: 51 passed
- Full test suite: 2217 passed, 1 skipped
- All commit hooks, including formatting, lint, and mypy, passed
- The cleaned Tamalpais config loads and the one-seed mission scan passes

### DIFF
--- a/conops/config/_base.py
+++ b/conops/config/_base.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel, ConfigDict, model_validator
+from pydantic import BaseModel, ConfigDict
 
 
 class ConfigModel(BaseModel):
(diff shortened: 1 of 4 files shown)


jak574 (approved):
lgtm

PR #229, opened 2026-07-29
feat: export plan-start osculating elements

Tohuvavohu:
## Problem

COAST exports TLE mean elements at the TLE epoch and a GCRS position/velocity timeseries, but it does not export instantaneous osculating elements for the propagated state. External trajectory comparisons therefore require consumers to repeat the Cartesian-to-COE conversion and can accidentally mix TLE mean elements with propagated elements or associate them with the wrong epoch.

## Changes

- add a typed `metadata.ephemeris.osculating_elements` block with an explicit epoch, `GCRS` frame, Earth-center origin, conversion note, and gravitational parameter
- derive the elements automatically from the exact simulation-start GCRS position and velocity when COAST attaches the orbit-state timeseries
- use rust-ephem for the Cartesian-state conversion; Coast only maps the result into its established SI plan-output keys
- leave central-body gravitational-parameter selection in rust-ephem and serialize the exact value it returns; rust-ephem currently defaults to the WGS-72 Earth value used by SGP4
- require the requested epoch to match exactly one ephemeris sample and require timestamps, positions, and velocities to have matching lengths
- preserve TLE mean elements and other ephemeris metadata regardless of attachment order
- document the distinction between `tle_mean_elements` and `osculating_elements`
- update the default-plan golden output to cover the new metadata contract

## Dependency

This PR is stacked on CosmicFrontierLabs/rust-ephem#180. Until that PR is merged and released, `pyproject.toml` points to its immutable commit `bc05341` so CI can exercise the real conversion. Before this PR is marked ready, replace that direct reference with the released rust-ephem version constraint.

## Validation

- full suite: `2221 passed, 1 skipped`
- focused metadata/schema/QueueDITL suite: `35 passed, 1 skipped`
- all configured hooks passed, including Ruff, pyupgrade, Sphinx, pyproject validation, and strict mypy
- default deterministic plan baseline passes with only the intentional osculating metadata addition
- smoke-tested against a real rust-ephem `TLEEphemeris` GCRS state and timestamp

### DIFF
--- a/conops/ditl/ditl_mixin.py
+++ b/conops/ditl/ditl_mixin.py
@@ -198,7 +198,11 @@ class DITLMixin:
 
     def _attach_orbit_state_timeseries_to_plan(self) -> None:
         """Attach GCRS spacecraft position/velocity samples to the current plan."""
-        from ..targets import OrbitStateSampleSchema, OrbitStateTimeseriesSchema
+        from ..targets import (
+            OrbitStateSampleSchema,
+            OrbitStateTimeseriesSchema,
+            attach_osculating_elements_metadata,
+        )
 
         pv = getattr(self.ephem, "gcrs_pv", None)
         positions = getattr(pv, "position", None)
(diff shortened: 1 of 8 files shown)


jak574 (approved):
lgtm

PR #230, opened 2026-07-30
fix: specify quaternion action in plan schema

Tohuvavohu:
## Problem

The plan schema identifies target attitude quaternions as `inertial_to_body` in `wxyz` order, but does not state the quaternion product or vector action. That is insufficient to distinguish COAST's Hamilton `q * v * conjugate(q)` action from consumers such as MAX that apply a frame quaternion as `conjugate(q) * v * q`.

This ambiguity caused MAX to apply COAST's numeric quaternion in the opposite sense. For ObsID 10483, the resulting boresight was about 7.9 degrees off and AD observed pervasive STR/imager occultations.

## Change

- Add machine-readable `quaternion_product: "hamilton"` and `vector_action: "q_v_q_conjugate"` fields to exported target attitudes.
- Document the exact vector transformation and the requirement for opposite-action consumers to conjugate the quaternion.
- Constrain the fields with literals and cover serialization/validation in tests.

This does not change any quaternion values or COAST pointing behavior. It makes the existing convention unambiguous so downstream format converters can translate it correctly.

## Validation

- `pytest`: 2217 passed, 1 skipped
- full pre-commit suite, including Sphinx and mypy
- default-plan golden regression

### DIFF
--- a/conops/targets/plan_entry.py
+++ b/conops/targets/plan_entry.py
@@ -26,12 +26,19 @@ BodyAxis = Literal["+X", "-X", "+Y", "-Y", "+Z", "-Z"]
 RollSource = Literal["planned", "defaulted_from_unconstrained_sentinel"]
 
 
-class AttitudeRotationSchema(BaseModel):
-    """Generic attitude rotation representation."""
+class AttitudeRotationConventionSchema(BaseModel):
+    """Machine-readable attitude rotation convention."""
 
     representation: Literal["quaternion"] = "quaternion"
     direction: Literal["inertial_to_body"] = "inertial_to_body"
     order: Literal["wxyz"] = "wxyz"
+    quaternion_product: Literal["hamilton"] = "hamilton"
+    vector_action: Literal["q_v_q_conjugate"] = "q_v_q_conjugate"
+
+
+class AttitudeRotationSchema(AttitudeRotationConventionSchema):
+    """Generic attitude rotation representation."""
+
     values: tuple[float, float, float, float]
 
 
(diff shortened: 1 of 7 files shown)


conops/targets/plan_entry.py
```diff
@@ -32,6 +32,8 @@ class AttitudeRotationSchema(BaseModel):
     representation: Literal["quaternion"] = "quaternion"
     direction: Literal["inertial_to_body"] = "inertial_to_body"
     order: Literal["wxyz"] = "wxyz"
+    quaternion_product: Literal["hamilton"] = "hamilton"
+    vector_action: Literal["q_v_q_conjugate"] = "q_v_q_conjugate"
```
jak574, line 36: Should these also be added to `AttitudeSampleSchema` in `plan.py`?

jak574 (changes requested):
One question about consistency.

conops/targets/plan_entry.py
Tohuvavohu: Yes. The executed samples come from the same `attitude_to_quat()` path and use the same convention. I added the convention to `AttitudeTimeseriesSchema` rather than repeating invariant metadata in every `AttitudeSampleSchema`: the sidecar now declares `GCRS` -> `COAST_BODY`, `wxyz`, Hamilton product, and `q_v_q_conjugate` once at the envelope level, matching the existing orbit-timeseries metadata pattern. Both the target rotation and attitude timeseries inherit the shared `AttitudeRotationConventionSchema` so they cannot drift. Docs, serialization tests, and the deterministic golden output are updated.

jak574 (approved):
lgtm

Issue #181, opened 2026-05-27
Support non-fixed antenna ground contacts without body-frame attitude tracking


Tohuvavohu:
Right now GSP attitude profile generation only supports fixed body-frame antennas. That is the right behavior for PR #177, because fixed antennas require the spacecraft body to point the configured antenna vector at the station.

But OMNI and GIMBALED antennas should not silently fall back to a fake fixed body vector. They need explicit handling:

- OMNI contacts should be schedulable without commanding spacecraft attitude to track the station.
- GIMBALED contacts should be modeled as steerable antenna contacts, with range/mode limits once that capability exists.
- Plan export and validation should make clear whether a contact is body-tracked or antenna-tracked.

The key thing: non-fixed antenna contacts should not imply legacy `-X` spacecraft tracking.

PR #231, opened 2026-07-30
feat: export external simulation initialization state

Tohuvavohu:
## Problem

External simulators may start before the observation plan so they can calibrate and settle. COAST exports TLE mean elements and a plan-window orbit-state timeseries, but neither provides an authoritative Cartesian state at an earlier initialization epoch. Consumers therefore have to repeat TLE propagation and can select a different gravity model, operation mode, or frame convention.

## Solution

- add typed `metadata.ephemeris.initialization_state` output
- include required UTC epoch, GCRS frame, Earth-center origin, SI position/velocity, propagator, gravity model, and SGP4 operation mode
- read propagation provenance directly from `TLEEphemeris`; callers cannot override source-owned metadata
- require the epoch to match exactly one ephemeris sample
- share exact-state selection with osculating metadata and reject misaligned timestamp/position/velocity arrays
- preserve TLE, osculating, and initialization metadata regardless of attachment order
- document the external-simulation handoff contract

The producer creates a dedicated ephemeris sample at the requested pre-plan epoch. This PR does not extend or alter the scheduler ephemeris.

## Dependency

This PR is stacked on #229 and pins rust-ephem commit `7a60b00` from rust-ephem #181. That stacked Rust commit includes #180 and supplies both the osculating conversion and source-owned TLE provenance API used here. The direct reference should return to the normal released dependency after those Rust PRs merge.

## Validation

- focused metadata/schema tests: 13 passed
- all repository hooks passed, including Ruff, Sphinx, workflow validation, and mypy
- full Tam plan generation completed with a 2,000-second pre-roll state
- exported state identifies `SGP4`, `WGS-72`, and `improved` operation mode

### DIFF
--- a/conops/targets/plan_metadata.py
+++ b/conops/targets/plan_metadata.py
@@ -149,15 +149,12 @@ def attach_osculating_elements_metadata(
     plan: Plan,
     ephemeris: rust_ephem.Ephemeris,
     epoch: datetime,
-    *,
-    mu_km3_s2: float = rust_ephem.WGS72_EARTH_MU_KM3_S2,
 ) -> None:
     """Attach GCRS osculating elements at an exact ephemeris timestamp.
 
-    ``mu_km3_s2`` defaults to the WGS-72 Earth value used by SGP4. Automatic
-    DITL plan export uses that default. Callers converting states generated
-    with another gravity model must pass its matching gravitational parameter;
-    the exact value used is included in the serialized elements.
+    rust-ephem owns the Cartesian-state conversion, including selection of its
+    central-body gravitational parameter. The exact value returned by
+    rust-ephem is included in the serialized elements.
     """
     timestamps = ephemeris.timestamp
     positions = ephemeris.gcrs_pv.position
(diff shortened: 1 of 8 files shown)


conops/targets/plan_metadata.py
```diff
@@ -204,3 +236,41 @@ def attach_osculating_elements_metadata(
...
         ),
     )
+
+
+def attach_initialization_state_metadata(
+    plan: Plan,
+    ephemeris: rust_ephem.TLEEphemeris,
+    epoch: datetime,
+) -> None:
+    """Attach one exact GCRS state for initializing an external simulation."""
```
jak574, line 246: So this code make the assumption that for a DITL the ephemeris will always be propagated from a TLE. For reasons I'm sure I don't have to tell you, this certainly will not be the case for all DITLs. I think a basic test here should check if this is a `TLEEphemeris`, I would suggest simplest would be inserting the following:
```python
    if not isinstance(ephemeris, rust_ephem.TLEEphemeris):
        return
```
We can think about how to generalize in the future.

jak574 (changes requested):

conops/targets/plan_metadata.py
Tohuvavohu: Addressed in 68e3844. The helper now accepts the generic Ephemeris type and returns without modifying plan metadata unless it receives a TLEEphemeris. Added a regression test for the non-TLE path.

jak574 (approved):

PR #232, opened 2026-07-31
fix: select reachable safe ground-pass profiles

Tohuvavohu:
## Problem

Ground-pass planning validates tracking attitudes over the contact window, but ingress safety depends on the spacecraft attitude immediately before the pass. The scheduler previously retried the preferred profile even when another complete, constraint-safe profile was reachable. At AOS, the fallback could also start PASS without confirming that ACS reached the validated tracking attitude.

Multi-seed validation exposed two related continuity failures:

- fixed-phase profiles were constraint-checked but not checked against ACS motion limits between tracking samples; the reproduced profile stepped 127.75 deg in 60 s (2.13 deg/s against a 2 deg/s limit)
- after reserving an accepted ingress/profile, later scheduler ticks could select another profile and mutate the pass before AOS

## Change

- retain the preferred pass-safe profile and reachable alternatives
- reject fixed-phase profiles that violate the configured ACS motion model between samples
- compute dynamic alternatives with one backward pass over the phase graph
- select ingress against the actual incoming attitude and PASS constraint scopes
- make an accepted pass reservation idempotent
- refuse to start PASS unless ACS reached the selected tracking attitude
- preserve compatibility for externally constructed `Pass` objects with only legacy start-attitude fields

## Validation

- pass and queue scheduler suites: 334 passed
- seed 0: 2 GSPs, 1,380 s PASS-mode time, 66,420 s science, zero mission-gate failures, 7.69 s generation
- combined Coast suite: 2,243 passed, 1 skipped
- default Coast plan output matches baseline
- 100-seed combined continuity scan: zero failed seeds, plan/execution mismatches, hard-constraint events, or adjacent-attitude rate violations; every seed executes 960 s of actual contact

This branch is based directly on `main`; the PR contains only the ground-pass scheduler change and its tests.

### DIFF
--- a/conops/ditl/queue_ditl.py
+++ b/conops/ditl/queue_ditl.py
@@ -1936,6 +1936,12 @@ class QueueDITL(DITLMixin, DITLStats):
         if next_pass is None:
             return False
 
+        # An accepted reservation already owns its ingress and tracking profile.
+        # Replanning it after the ingress slew completes can select a different
+        # profile and leave the spacecraft off-attitude at contact start.
+        if self._ground_pass_key(next_pass) in self._planned_gsp_keys:
+            return False
+
         # Check if it's time to start slewing for the next pass. Tracking
         # profiles are all pass-safe, but their incoming slew paths depend on
         # the actual attitude at this step.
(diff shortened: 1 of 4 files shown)


jak574 (approved):
This looks fine, but please merge the predecessor PRs before this one.

PR #233, opened 2026-07-31
fix: reject attitude telemetry that exceeds slew-rate limit

Tohuvavohu:
## Problem

Plan/execution validation checked task matching and attitude constraints, but did not check kinematic continuity between adjacent exported attitude samples. A mode transition could therefore replace the executed attitude with a distant attitude and still export a plan, effectively allowing a spacecraft attitude teleport.

## Solution

- validate the exact `Telemetry.housekeeping` stream used to write `attitude_timeseries.json`
- compute the shortest full SO(3) quaternion rotation between every adjacent RA/Dec/roll sample
- require `rotation_deg <= max_slew_rate_deg_per_s * elapsed_seconds` with no ACS-mode exemptions
- reject non-increasing/non-finite timestamps and non-finite attitudes
- run the invariant in both DITL implementations before sidecar export
- report sample indices, timestamps, modes, actual rotation/rate, and configured limit

This is a sampled-telemetry invariant. It complements command-boundary transition guards; it cannot prove what occurs between two samples when a discontinuity is smaller than the allowed motion over that whole interval.

## Validation

- `pytest -q`: 2,225 passed, 1 skipped
- focused DITL/QueueDITL tests: 305 passed
- all hooks pass, including Ruff and mypy
- default Coast plan output matches the golden baseline
- combined with #232, #234, and #235: 2,243 passed, 1 skipped; 100 Tamalpais seeds have zero failed seeds, plan/execution mismatches, hard-constraint events, and adjacent-attitude rate violations

The gate exposed three pre-existing mission intervals at 2.655, 2.167, and 2.129 deg/s against a 2.0 deg/s limit. #232 fixes internal pass tracking, #234 preserves the pass-end attitude, and #235 constrains continuous roll optimization to reachable candidates.

### DIFF
--- a/conops/ditl/ditl.py
+++ b/conops/ditl/ditl.py
@@ -134,7 +134,10 @@ class DITL(DITLMixin, DITLStats):
                 (missing ephemeris, missing plan, or invalid ephemeris date range).
 
         Raises:
-            No exceptions raised; errors are logged to stdout and return False.
+            ValueError: If the ephemeris or plan is missing, or the ephemeris does
+                not cover the simulation date range.
+            AttitudeRateContinuityError: If adjacent executed attitude samples
+                exceed the configured maximum slew rate.
 
         Note:
             The simulation respects the class attributes:
@@ -168,7 +171,6 @@ class DITL(DITLMixin, DITLStats):
         simlen = len(self.utime)
         self.ra = np.zeros(simlen).tolist()
         self.dec = np.zeros(simlen).tolist()
-        self.roll = np.zeros(simlen).tolist()
         self.mode = np.zeros(simlen).astype(int).tolist()
         self.panel = np.zeros(simlen).tolist()
         self.obsid = np.zeros(simlen).astype(int).tolist()
@@ -223,7 +225,6 @@ class DITL(DITLMixin, DITLStats):
             self.batteryalert[i] = self.battery.battery_alert
             self.ra[i] = ra
             self.dec[i] = dec
-            self.roll[i] = roll
             self.mode[i] = mode
             self.panel[i] = panel_illumination
             self.power[i] = power_usage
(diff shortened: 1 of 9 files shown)


conops/ditl/ditl_mixin.py
```diff
@@ -13,6 +16,50 @@
 from ..targets import Plan, PlanEntry
 from .telemetry import Telemetry
 
+ATTITUDE_RATE_NUMERICAL_TOLERANCE_DEG = 1e-9
+
+
+@dataclass(frozen=True)
+class _AttitudeRateViolation:
```
jak574: Why use a dataclass here where we're using Pydantic elsewhere?

conops/ditl/ditl.py
```diff
@@ -374,6 +377,7 @@ def calc(self) -> bool:
                 )
                 self.telemetry.data.append(pd)
 
+        self._assert_attitude_rate_continuity()
```
jak574, line 380: Why isn't this called in `queue_ditl.py`?

jak574 (changes requested):

conops/ditl/ditl_mixin.py
Tohuvavohu: Changed in a95b0b0. AttitudeRateViolation now follows the repository frozen Pydantic BaseModel pattern.

conops/ditl/ditl.py
Tohuvavohu: It was enforced indirectly because QueueDITL.calc called the plan-execution assertion, whose validator included attitude-rate violations. I agree that was unclear at the call site, so a95b0b0 now calls the dedicated continuity assertion explicitly before plan/execution validation and sidecar export. The validation API still reports the same violation details.

jak574 (approved):
lgtm

PR #234, opened 2026-07-31
fix: preserve attitude continuity after ground passes

Tohuvavohu:
## Problem

Ending a ground-station pass cleared the stale ingress slew without preserving the final tracked attitude. The next ACS update could recompute roll before the following slew started, producing an instantaneous attitude jump. In the Tamalpais reproduction, the pass-exit sample moved 159.28 deg in 60 s (2.65 deg/s against a 2 deg/s limit).

## Change

- Replace stale/precomputed slew state at `END_PASS` with a zero-duration IDLE hold at the executed pass-end RA/Dec/roll.
- Keep future slew commands queued; `_start_slew()` still recalculates their start from the current physical attitude when they execute.
- Correct the pass-end log so it reports the hold instead of claiming an uncommanded return to the last PPT.

## Plan impact

Target selection after a pass now starts from the real pass-end roll instead of the previous discontinuous roll. The Tamalpais target sequence therefore changes. In the 100-seed combined safety run, science ranges from 64,080 to 65,700 s, so tamalpais-configuration#37 rebaselines the old 65,400 s scanner floor separately.

## Validation

- `pytest -q tests/acs/test_acs.py tests/acs/test_acs_coverage.py tests/scheduler_queue/test_queue_ditl.py` (368 passed)
- Combined continuity integration: 2,243 passed, 1 skipped
- Default Coast plan output matches baseline
- 100-seed Tamalpais integration with #232, #233, and #235: zero failed seeds, plan/execution mismatches, hard-constraint events, and adjacent-attitude rate violations; generation is 6.73-8.99 s.

### DIFF
--- a/conops/simulation/acs.py
+++ b/conops/simulation/acs.py
@@ -283,17 +283,6 @@ class ACS:
         self.current_pass = None
         self.acsmode = ACSMode.IDLE
 
-        # Clear any stale slew that was issued during the pass. Such slews have
-        # start positions from pass tracking that don't reflect where the
-        # spacecraft actually is now. Keeping them would cause teleportation.
-        if self.last_slew is not None and self.last_slew.slewstart < utime:
-            self._log_or_print(
-                utime,
-                "PASS",
-                f"{unixtime2date(utime)}: Clearing stale slew (started {self.last_slew.slewstart:.0f})",
-            )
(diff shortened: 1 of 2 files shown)


jak574 (approved):
lgtm

PR #235, opened 2026-07-31
fix: keep optimized roll within ACS motion limits

Tohuvavohu:
## Problem

Charging and safe-mode dwell recomputed the globally optimal constraint-safe roll at every telemetry sample without considering the previously executed roll. When the best valid interval moved, ACS could jump between valid intervals. The Tamalpais reproduction jumped 130 deg in 60 s (2.17 deg/s against a 2 deg/s limit), although a nearby safe roll was available.

## Change

- Add `AttitudeControlSystem.max_motion_angle()` as the inverse of the existing rest-to-rest bang-bang motion model.
- Allow `optimum_roll()` to restrict candidates to a shortest-path window around the executed roll.
- During continuous charging/safe roll optimization, search only sample-valid rolls reachable in the elapsed dwell time. The first update after another controller owned attitude has a zero-degree window.
- If no integer-degree valid candidate is reachable, hold the current roll instead of discontinuously jumping; normal constraint validation can then reject an infeasible plan.

For Tamalpais limits (2 deg/s, 0.125 deg/s²), a 60 s dwell update may move at most 88 deg under the configured acceleration/cruise/deceleration model. The reproduced case selects the nearby 209 deg safe roll instead of jumping from 215 deg to 345 deg.

This enforces endpoint reachability at the telemetry cadence. Continuous keepout safety between samples remains the responsibility of slew-path/constraint validation; this PR does not claim a sub-sample proof.

## Validation

- Focused ACS/roll/config/scheduler tests: 471 passed
- Combined continuity integration: 2,243 passed, 1 skipped
- Default Coast plan output matches baseline
- 100-seed Tamalpais integration with #232, #233, and #234: zero failed seeds, plan/execution mismatches, hard-constraint events, and adjacent-attitude rate violations; generation is 6.73-8.99 s.

### DIFF
--- a/conops/config/acs.py
+++ b/conops/config/acs.py
@@ -80,6 +80,19 @@ class AttitudeControlSystem(ConfigModel):
         t_cruise = d_cruise / vmax
         return float(2 * t_accel + t_cruise)
 
+    def max_motion_angle(self, duration_s: float) -> float:
+        """Maximum rest-to-rest angular motion in the available time."""
+        if duration_s <= 0:
+            return 0.0
+        a = float(self.slew_acceleration)
+        vmax = float(self.max_slew_rate)
+        if a <= 0 or vmax <= 0:
+            return 0.0
+        t_accel = vmax / a
+        if duration_s <= 2 * t_accel:
+            return float(0.25 * a * duration_s**2)
+        return float(vmax * (duration_s - t_accel))
+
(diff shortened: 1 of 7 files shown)


jak574 (approved):
lgtm

jak574 (commented):
lgtm

PR #239, opened 2026-08-04
fix: osculating element model integration

Tohuvavohu:
## Problem

rust-ephem now returns `OsculatingElements` as a Pydantic model, but COAST still indexes the result as a dictionary. With current merged versions, `QueueDITL.calc()` fails while attaching orbit metadata:

```
TypeError: 'OsculatingElements' object is not subscriptable
```

The existing rust-ephem dependency pin points to the pre-review implementation that returned a `TypedDict`, so CI did not exercise the merged API.

## Solution

- require the released `rust-ephem>=0.13.0` package
- read the documented `OsculatingElements` model attributes directly
- keep the serialized COAST plan metadata schema unchanged

## Validation

- `pytest tests/plan/test_plan_metadata.py -q`: 14 passed
- `prek run --all-files`: passed
- full 24-hour plan generation: passed, 90 entries
- repeatability check: passed
- seed-0 mission scan: passed with zero hard-constraint events and zero validation/constraint mismatches

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "pydantic>=2.12",
     "pyproj>=3.7.0",
     "pytz>=2025.2",
-    "rust-ephem @ git+https://github.com/CosmicFrontierLabs/rust-ephem.git@586eeff",
+    "rust-ephem>=0.13.0",
     "shapely>=2",
     "tqdm>=4.67",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "pydantic>=2.12",
     "pyproj>=3.7.0",
     "pytz>=2025.2",
-    "rust-ephem @ git+https://github.com/CosmicFrontierLabs/rust-ephem.git@7a60b00",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -36,7 +36,7 @@ dependencies = [
     "pydantic>=2.12",
     "pyproj>=3.7.0",
     "pytz>=2025.2",
-    "rust-ephem @ git+https://github.com/CosmicFrontierLabs/rust-ephem.git@7a60b00",
+    "rust-ephem @ git+https://github.com/CosmicFrontierLabs/rust-ephem.git@586eeff",
```
jak574: If this is relying on a unreleased `rust-ephem`, let's get it released and pin it properly. I don't like fixing to a github install by default.

jak574 (changes requested):
lgtm, but need to fix pyproject.toml issue.

pyproject.toml
Tohuvavohu: Updated in 4427c1d to use `rust-ephem>=0.13.0`. The v0.13.0 release has now completed and is published to PyPI.

jak574 (approved):
lgtm

PR #240, opened 2026-08-11
feat: adopt right-handed spacecraft roll convention

Tohuvavohu:
## Problem

COAST historically exposed roll with the sign of its passive inertial-to-body transform, while rust-ephem defines spacecraft roll as a physical right-handed rotation about body +X. Keeping an adapter at one API boundary would leave two meanings for the same public quantity and make every future caller responsible for knowing which convention it had.

rust-ephem #189 and #191 fixed coordinated roll composition across roll sweeps, fixed-roll evaluation, compound trees, and polygon constraints. COAST should use that convention directly everywhere.

## Solution

- define positive COAST roll as a right-handed physical spacecraft rotation about body +X
- keep inertial-to-body quaternions mathematically correct by applying the inverse `R_x(-roll)` transform
- define zero roll by aligning body +Z with projected celestial north, with the RA/Dec Euler sequence providing the pole continuation
- update attitude conversion, extraction, roll optimization, star-tracker transforms, and COAST visualization helpers to the same sign
- pass scalar, batch, and roll-range values directly to rust-ephem without boundary adapters
- export the complete convention at plan, target-attitude, and attitude-timeseries scope: `roll_axis`, `roll_convention`, `roll_reference_axis`, and `roll_reference`
- require rust-ephem 0.14.2, which contains both upstream roll fixes
- make dynamic GSP path ordering insensitive to sub-nanodegree floating-point noise exposed by the representation migration

This intentionally changes the sign of public numeric roll values. At RA=0 and Dec=0, +90 degrees places body +Y along inertial +Z. Exported quaternions still represent the same physical attitudes; their component arrays may differ by the equivalent global sign `q` versus `-q`.

## Validation

- full COAST suite against the official PyPI rust-ephem 0.14.2 wheel: 2,267 passed, 1 skipped
- all repository hooks passed, including Ruff, mypy, Sphinx, and workflow validation
- scalar and batch quaternion round trips preserve physical attitude at both celestial poles
- the two-star-tracker roll-range mask matches independent fixed-roll evaluation at all 360 integer-degree candidates
- the default-plan baseline was regenerated and independently checked for zero physical-attitude or non-representation differences
- full Tamalpais seed-0 generation with the official 0.14.2 wheel produced 90 entries, 67,518 seconds of science, two GSPs, and 960 seconds of contact
- structured old-adapter versus canonical-convention comparison found zero plan, attitude-timeseries, or orbit-state differences after applying only `old_roll = -new_roll mod 360` and treating `q` and `-q` as equivalent

## Dependency

rust-ephem #189 and #191 are included in rust-ephem 0.14.2, published by release PR #192. The official PyPI wheel was used for the full COAST suite and full-plan validation above.


jak574 (approved):
lgtm

PR #243, opened 2026-08-13
fix: align boresight pitch convention with rust-ephem

### LINKED ISSUE #241: Align non-planar boresight pitch convention with rust-ephem
## Problem

COAST converts a body boresight with a positive +Z component to a positive pitch in both `normal_to_euler_deg()` and `StarTrackerConfiguration._boresight_to_euler_deg()`. rust-ephem's boresight geometry defines the local Z component as `-sin(pitch)`, so the same positive pitch rotates toward -Z.

Physical fixed-roll parity therefore fails for +Z/-Z and generic boresights with a nonzero Z component. Pure +/-Y boresights are unaffected because their pitch is zero. This was found while auditing coast-sim#240 and is independent of the roll-sign fix there.

## Acceptance criteria

- define and document one pitch/yaw convention at the COAST/rust-ephem boundary
- make +Z, -Z, and generic non-planar boresights map to the intended physical axes
- add end-to-end parity tests comparing COAST body-vector geometry with rust-ephem fixed constraint evaluation
- keep the conversion centralized rather than applying per-call sign fixes

Tohuvavohu:
## Problem

COAST converted a body normal with a positive Z component to positive boresight pitch. rust-ephem's intrinsic Z-Y-X convention maps positive pitch toward local -Z, so planning constraints mirrored every non-planar mounted boresight across the body XY plane. Pure +/-Y mounts were unaffected, which hid the mismatch in the existing star-tracker parity coverage.

## Solution

- define the rust-ephem boresight-offset convention in the explicitly named `normal_to_boresight_offset_euler_deg()` boundary helper
- map positive body Z to negative rust-ephem pitch
- remove the duplicate star-tracker conversion and use the same helper for star trackers, radiators, and telescopes
- preserve explicit +X, +/-Y, +/-Z, XY, XZ, and generic non-planar mapping coverage
- add fixed-roll parity coverage for +Z, -Z, and generic non-planar boresights

The parity regression compares COAST's direct body-vector transform with rust-ephem's actual wrapped constraint evaluation across multiple attitudes and spacecraft rolls.

## Verification

- `2266 passed, 1 skipped`
- pre-commit hooks pass for all changed files
- GitHub CI passes on Python 3.10 through 3.14, plus lint, formatting, and type checking

Fixes #241

### DIFF
--- a/conops/config/radiator.py
+++ b/conops/config/radiator.py
@@ -19,7 +19,7 @@ from pydantic import Field, field_validator
 from rust_ephem.constraints import ConstraintConfig
 
 from ..common import dtutcfromtimestamp, scbodyvector
-from ..common.vector import normal_to_euler_deg, radec2vec
+from ..common.vector import normal_to_boresight_offset_euler_deg, radec2vec
 from ._base import ConfigModel
 from .constraint import Constraint
 from .geometry import PanelGeometry, compute_shadow_fraction
@@ -130,7 +130,9 @@ class Radiator(ConfigModel):
         if base_constraint is None:
             return None
 
-        roll_deg, pitch_deg, yaw_deg = normal_to_euler_deg(self.orientation.normal)
+        roll_deg, pitch_deg, yaw_deg = normal_to_boresight_offset_euler_deg(
+            self.orientation.normal
+        )
         return base_constraint.boresight_offset(
             roll_deg=roll_deg,
             pitch_deg=pitch_deg,
--- a/conops/config/radiator.py
+++ b/conops/config/radiator.py
@@ -129,6 +129,8 @@ class Radiator(ConfigModel):
         base_constraint = self.hard_constraint.constraint
         if base_constraint is None:
             return None
+        if self.orientation.normal == (1.0, 0.0, 0.0):
+            return base_constraint
 
         roll_deg, pitch_deg, yaw_deg = normal_to_boresight_offset_euler_deg(
             self.orientation.normal
--- a/conops/config/star_tracker.py
+++ b/conops/config/star_tracker.py
@@ -38,7 +38,13 @@ from rust_ephem import AtLeastConstraint, EarthLimbConstraint, SunConstraint
 from rust_ephem.constraints import ConstraintConfig
 
 from ..common.enums import ACSMode
-from ..common.vector import radec2vec, rotvec, vec2radec, vecnorm
+from ..common.vector import (
+    normal_to_boresight_offset_euler_deg,
+    radec2vec,
+    rotvec,
+    vec2radec,
+    vecnorm,
+)
 from ._base import ConfigModel
 from .constraint import Constraint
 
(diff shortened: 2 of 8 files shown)


conops/config/radiator.py
```diff
@@ -130,7 +130,9 @@ def _hard_constraint_with_offset(self) -> ConstraintConfig | None:
         if base_constraint is None:
             return None
```
jak574, line 131: Looking at this code, I think in the case where the constraint is aligned with the regular boresight, we shouldn't apply the offset. So adding something like this here:

```python
        if self.orientation.normal == (1.0, 0.0, 0.0):
            return base_constraint
```

conops/config/star_tracker.py
jak574, line 87: Should pitch_rad here have a negative applied given the new convention?

conops/config/star_tracker.py
```diff
@@ -396,7 +380,7 @@ def startracker_hard_constraint(self) -> ConstraintConfig | None:
             if base_constraint is None:
                 continue
```
jak574, line 374: Again here, probably should just pass through constraint if boresight of tracker is aligned with pointing direction, but here because we're possibly combining with another constraint, you can't just return here...
```python
            if st.orientation.boresight == (1.0, 0.0, 0.0):
                offset_constraint = base_constraint
            else ...
```

jak574 (changes requested):
Some things that should make the code more efficient, plus one math question.

conops/config/radiator.py
Tohuvavohu: Fixed in 6e91d50. A +X radiator normal now returns the base constraint directly, with regression coverage that verifies no boresight-offset wrapper is created.

conops/config/star_tracker.py
Tohuvavohu: No. This helper defines COAST body-frame mounting angles: positive pitch rotates +X toward body +Z. rust-ephem boresight-offset Euler angles use the opposite pitch sign for that same vector, and normal_to_boresight_offset_euler_deg() owns that API-boundary conversion. Negating pitch here would reverse existing configured tracker mounting directions. I clarified the docstring and added a regression test showing +45 deg body pitch -> +Z boresight -> -45 deg rust-ephem offset pitch.

conops/config/star_tracker.py
Tohuvavohu: Fixed in 6e91d50. Aligned +X trackers now contribute the base constraint directly; only off-axis trackers are wrapped. I applied the same rule to both hard and soft constraint aggregation and added coverage for each path.

conops/config/star_tracker.py
jak574: I believe that the _zero_roll_body_axes helper function should be used here also.
```suggestion
        # Spacecraft boresight (+X) and the zero-roll Y/Z basis in ECI.
        x_hat, y0, z0 = _zero_roll_body_axes(np.deg2rad(ra_deg), np.deg2rad(dec_deg))
```

jak574 (changes requested):

conops/config/star_tracker.py
Tohuvavohu: Fixed in ed9cf56. transform_pointing() now uses _zero_roll_body_axes() directly. I also added a celestial-pole regression test, where the shared analytical basis preserves the RA-defined zero-roll convention instead of switching to the old arbitrary fallback axis.

jak574 (approved):
lgtm

PR #245, opened 2026-08-14
fix: preserve compound roll constraint semantics

### LINKED ISSUE #242: Preserve AtLeast semantics in roll-dependent constraint extraction
## Problem

`Constraint.roll_dependent_constraint` recursively collects every `BoresightOffsetConstraint` leaf and OR-combines the flattened list. Flattening changes the semantics of compound trees, especially `AtLeastConstraint`.

For example, a two-tracker `AtLeast` tree that permits one tracker to be constrained should reject a roll only when both tracker leaves are violated. The flattened OR rejects whenever either tracker is violated, making the roll mask substantially over-restrictive. This was found while auditing coast-sim#240 and is independent of its roll-sign change.

## Acceptance criteria

- preserve AND/OR/NOT/XOR/AtLeast structure when projecting a constraint tree to its roll-dependent portion
- define how roll-independent children are reduced without changing thresholds
- compare the optimized roll mask against fixed-roll evaluation for multiple AtLeast thresholds
- include a two-tracker regression where one functional tracker is sufficient

Tohuvavohu:
## Problem

`Constraint.roll_dependent_constraint` flattened every nested
`BoresightOffsetConstraint` into one OR tree. That changed the meaning of
compound constraints. In particular, a two-tracker `AtLeast(min_violated=2)`
tree became "reject when either tracker is constrained" instead of "reject
when both trackers are constrained."

## Solution

- detect boresight-offset descendants without flattening their parent tree
- omit a tree only when it is entirely roll-independent
- retain a complete mixed tree whenever any descendant depends on roll
- let rust-ephem's compound `roll_range()` evaluate roll-independent siblings as
  constants for the current target and time

Retaining mixed trees is required for exact semantics. A roll-independent child
can be either violated or clear at runtime, which changes AND, OR, NOT, XOR, and
AtLeast results. It therefore cannot be removed by static extraction. AtLeast
thresholds and all parent structure remain unchanged.

## Regression coverage

The optimized mask is compared with fixed-roll evaluation across all 360 integer
rolls for:

- two trackers with one functional tracker required (`min_violated=2`)
- two trackers with both functional trackers required (`min_violated=1`)
- a mixed AtLeast tree whose roll-independent child is clear
- the same mixed tree whose roll-independent child is violated

The old flattened implementation allows only 242 of 360 rolls in the first
two-tracker case. The earlier static-reduction approach also produced 118 mask
mismatches when the mixed tree's independent child was violated.

## Verification

- full COAST suite on current `main`: `2273 passed, 1 skipped`
- lint, formatting, and strict mypy checks passed
- Tamalpais seed 0 passes mission gates
- detailed seed statistics remain unchanged: 90 entries, 65,100 science
  seconds, 960 contact seconds, 11,220 slew seconds, and zero hard-constraint or
  validation-mismatch events

Closes #242

### DIFF
--- a/conops/config/constraint.py
+++ b/conops/config/constraint.py
@@ -6,7 +6,15 @@ from typing import cast
 import numpy as np
 import rust_ephem
 from pydantic import ConfigDict, Field, PrivateAttr
-from rust_ephem.constraints import ConstraintConfig
+from rust_ephem.constraints import (
+    AndConstraint,
+    AtLeastConstraint,
+    BoresightOffsetConstraint,
+    ConstraintConfig,
+    NotConstraint,
+    OrConstraint,
+    XorConstraint,
+)
 
 from ..common import dtutcfromtimestamp
 from ..common.enums import ACSMode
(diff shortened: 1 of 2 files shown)


jak574 (approved):
lgtm

